PR #5, opened 2020-08-07
Adding project shell for backend

gunnarmorling:
Hey @mdrillin, here's the initial import of the backend. It doesn't doo much yet apart from serving the frontend via the Quarkus server. Next I'm going to add some basic REST endpoint for you to invoke from the client.

The build is described in the README, give it a try and let me know whether it works. Note I still need to do an explicit `yarn build` in order to have UI changes to be picked up by the backend. Not quite sure why this is, as it was my assumption that the Maven frontend plug-in would take care of this automatically. I reckon things are cached somewhere. If you got any idea around that, I'll be very happy to learn about it :)


gunnarmorling:
// CC @rk3rn3r @Naros @jpechane

PR #7, opened 2020-08-07
Adding /api/connector endpoint for returning all available connector …

gunnarmorling:
…types;

Returning a static list for now to get going.

@mdrillin, here's a quick mock endpoint which returns all "available" connector types. You might use this to get REST API calls going and drive the initial connector selection page (there's an "enabled" flag which only is true for the PG connector). Is that useful for you to move on?


gunnarmorling:
Note you can start the backend in dev mode (see README) and then you should be able to point the client in the yarn dev mode to it, so you don't need to deploy everything via Quarkus if you're working on the UI.

gunnarmorling:
@jcechace, now with that basic infra in place, you also might want to take a first look, so to see whether you can sprinkle some QE pixie dust and make sure we start off into the right direction?

mdrillin:
@gunnarmorling yeah this is definitely useful - @indraraj has the connector selection component in place, so we can work on hooking that up to the rest call now.  Thanks!

gunnarmorling:
Cool, merged it.

PR #10, opened 2020-08-11
Initial backend

gunnarmorling:
Hey @mdrillin, expanded the API some more. There's now a (very basic) method for validating connection properties.


gunnarmorling:
// CC @rk3rn3r.

gunnarmorling:
This still all is very much in flux; I'll add the Swagger UI, so you can easily get an overview over the currently exposed methods.

gunnarmorling:
👋  @mdrillin, @indraraj, @uidoyen, you can get API specs now from http://localhost:8080/ and http://localhost:8080/swagger-ui.

gunnarmorling:
@mdrillin, added one more endpoint, _/api/connector-types/{id}/validation/properties_ for validating an arbitrary set of options. My thinking being, that you would invoke

* _/connector-types/{id}/validation/connection_ on the page/tab with the connection properties
* _/connector-types/{id}/validation/filters_ on the page with the include/exclude expressions; that's still tbd., and it'd yield the set of tables matching the passed filter(s)
* _/connector-types/{id}/validation/properties_ on any other page, passing the fields on that one

Does that compute?

mdrillin:
@gunnarmorling yeah, that makes sense.  I noticed the endpoint in ConnectorResourceTest needs to be updated to 'connector-types', but looks like a great start.

gunnarmorling:
Whoops, good catch with the tests. Will fix.

gunnarmorling:
Ok, test fixed and also added the filter endpoint.

With this, I think you should be able to implement all the wizard steps, apart from the very last one for actually putting the assembled configuration in place. Note the server is stateless, i.e. the assumption is that the client itself will propagate the configuration from one page to the next before sending the complete configuration at the end.

I'm going to merge this one now, and will next look into adding that last missing API method; or perhaps @rk3rn3r might take over, too. Note you need to have Debezium 1.3.0-SNAPSHOT built and installed locally based on PR https://github.com/debezium/debezium/pull/1752, as I had to do some small changes in Debezium Core, too. To do so, check out this branch (or the master branch once that PR has been merged) and build it like so:

    mvn clean install -pl :debezium-core,:debezium-connector-postgres -DskipTests=true -DskipITs=true -Dcheckstyle.skip=true

Let me know how it goes :)

gunnarmorling:
Btw. I saw you did some UI changes, too. Is this about calling the backend API already?

mdrillin:
> @mdrillin, btw. I saw you did some UI changes, too. Is this about calling the backend API already?

@gunnarmorling yes, I verified the UI could connect to the backend and so the first page will show the backend data.  @indraraj has PR'd some additional framework changes - we will work to get that merged today.

gunnarmorling:
Awesome, thanks a lot! Bilgin set up a sync-up call for tomorrow, perhaps
you could demo then how things look like atm. (if doable, no problem if
not).

>

PR #22, opened 2020-08-16
Ignoring node_modules and Implementing postCSS

uidoyen:
Ignoring node_modules from root folder and implementing postCSS to use sass


indraraj:
thanks, @uidoyen changes look good. I think we can let the yarn.lock file committed. https://classic.yarnpkg.com/en/docs/yarn-lock/#toc-managed-by-yarn

uidoyen:
> I think we can let the yarn.lock file committed. https://classic.yarnpkg.com/en/docs/yarn-lock/#toc-managed-by-yarn

Just removing from root folder where we don't have any node dependencies.

indraraj:
sure, it good to go then   :+1:

PR #26, opened 2020-08-19
Add "create connector" endpoint

rk3rn3r:
Add "create connector" enpoint

### DIFF
--- a/backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
+++ b/backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
@@ -1,5 +1,6 @@
 package io.debezium.configserver.rest;
 
+import java.io.IOException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.Arrays;
@@ -20,6 +21,7 @@ import javax.ws.rs.GET;
 import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
+import javax.ws.rs.ProcessingException;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.Response;
@@ -38,14 +40,14 @@ import io.debezium.configserver.service.ConnectorIntegrator;
 @Path("/api")
 public class ConnectorResource {
 
+    public static final String PROPERTY_KAFKA_CONNECT_URI = "kafka.connect.uri";
+
     private static final Logger LOG = Logger.getLogger(ConnectorResource.class);
 
-    @ConfigProperty(name = "kafka.connect.uri", defaultValue = "http://localhost:9000")
-    private String kafkaConnectBaseUri = "http://localhost:9000";
+    @ConfigProperty(name = PROPERTY_KAFKA_CONNECT_URI, defaultValue = "http://localhost:8083")
+    String kafkaConnectBaseUri = "http://localhost:8083";
 
     private final Map<String, ConnectorIntegrator> integrators;
-
-    private final URI kafkaConnectURI;
 
     public ConnectorResource() {
         Map<String, ConnectorIntegrator> integrators = new HashMap<>();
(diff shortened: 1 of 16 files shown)


backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
```diff
@@ -23,13 +32,21 @@
...
 
 @Path("/api")
 public class ConnectorResource {
 
+    public static final String PROPERTY_KAFKA_CONNECT_URI = "kafka.connect.uri";
+
+    private static final Logger LOG = Logger.getLogger(ConnectorResource.class);
+
+    @ConfigProperty(name = PROPERTY_KAFKA_CONNECT_URI, defaultValue = "http://localhost:8083")
+    String kafkaConnectBaseUri = "http://localhost:8083";
```
gunnarmorling, line 48: I think this can be injected as a `List<URI`>.

gunnarmorling (commented):
LGTM, thanks, @rk3rn3r!

As discussed, by configuring the KC cluster URLs in the backend, we'll need to have an endpoint for exposing the list of known clusters to the UI.

// CC @mdrillin Let's talk about this in the call later today. The idea is to configure the list of Kafka Connect clusters (i.e. the thing we want to configure here) in the UI backend server, and then it'd be a matter of showing the list of available clusters to the UI user and having them select the one to use when configuring a connector.

backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
rk3rn3r: `Caused by: java.lang.RuntimeException: Error injecting java.util.List&lt;java.net.URI&gt; io.debezium.configserver.rest.ConnectorResource.kafkaConnectBaseUris` when using a comma-separated list.

rk3rn3r: Should we allow whitespaces in the URI list? then we need to to it manually, but we can do `List<String>`. @gunnarmorling

gunnarmorling: Let's disallow whitespaces then. Worth double-checking on quarkus-dev though whether that's a known limitation for List<URI> or whether it's a bug we should log.

backend/src/test/java/io/debezium/configserver/util/PostgresInfrastructureTestProfile.java
```diff
@@ -0,0 +1,18 @@
...
+import io.quarkus.test.junit.QuarkusTestProfile;
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+public class PostgresInfrastructureTestProfile implements QuarkusTestProfile {
+    @Override
+    public List<TestResourceEntry> testResources() {
+        ArrayList<TestResourceEntry> resourceEntries = new ArrayList<>(1) {};
```
gunnarmorling, line 12: Any specific reason for not using the much shorter `Arrays.asList(...)`?

rk3rn3r: just the way how it gathered... will use `Collections.singletonList(...)`

backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
rk3rn3r: Not really a bug. `List<>` injection from a config property will split on comma. And then it leaves all fields untouched. but a URI with `    http://localhost:123    ` is just wrong because it's not allowed to start or end URI with whitespaces.
I addressed that in the last or second last commit with making it a `List<String>`. Can you live with that? Imo it makes the config more accessible to users.

gunnarmorling: Sure, works for me. Still not sure though why Quarkus couldn't trim comma-separated URI values?

gunnarmorling:
Thx!

@mdrillin, so here's the endpoint for registering connectors :)

backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
rk3rn3r: I guess there's just no specific handling for `List<URI>`, but that's just a guess. It's risky to manipulate the input and so they don't trim it, also it is safe for URI.

rk3rn3r:
@mdrillin aaand the `/api/connect-clusters` endpoint as well!

PR #34, opened 2020-08-21
use connector id to identify

mdrillin:
I missed that there was an id on the ConnectorType in my initial commit.  Better to identify ConnectorType by the id vs the longer classname - so I added the id to the model, then converted the usages.  Also added some comments in ConnectorService


indraraj (approved):
looks good.

PR #35, opened 2020-08-24
Adding routing

indraraj:
- Added the router config 

- integrated the PageNotFound component.

- added the react-router and react-router-dom to no-hoist for using react-router-dom hooks

- Added some basic SelectTablesStep Component.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -10,7 +10,9 @@
     "nohoist": [
       "**/@patternfly/react-core",
       "**/@patternfly/react-core/**",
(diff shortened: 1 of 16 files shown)


indraraj:
> Looks like a good start to me - we can adjust in subsequent PRs if needed. I would probably rename 'SelectTables..' component to something like 'DefineFilters..' but I'm good to merge.

@mdrillin  my thoughts behind using the name 'SelectTables...' was to have the UI component name resemble the UI feature it represents on screen.

PR #39, opened 2020-08-25
fixing build error

uidoyen:
Fixing build error due to connector properties form.


indraraj (approved):
Tested locally, looks good.

PR #44, opened 2020-08-27
Supply ConnectorProperties to components

mdrillin:
Some initial work on supplying the correct subset of ConnectorProperties to the relevant pages.
- added ConnectorConfiguration model.  This is the object that will be passed to the 'createConnector' service
- added optional 'category' attribute to ConnectorProperty (this is not used yet)
- added createConnector function to the service, but it's commented out/not used yet.
- the FiltersStepComponent/ConfigureConnectorTypeComponent have properties for setting the relevant ConnectorProperties.
- added functions in Utils which group the property definitions by 'category' for supplying to the components.  This will change once the backend(hopefully) is able to supply category info.
- Each component has 'onSaveProperties' and 'onValidateProperties'.  The config property name-value pairs can then be supplied, and the CreateConnectorPage is responsible for doing the save or validation.

### DIFF
--- a/ui/packages/ui/src/app/shared/Utils.ts
+++ b/ui/packages/ui/src/app/shared/Utils.ts
@@ -1,10 +1,17 @@
-import { ConnectorType } from "@debezium/ui-models";
+import { ConnectorConfiguration, ConnectorType } from "@debezium/ui-models";
+import { ConnectorProperty } from '@debezium/ui-models';
 
 export enum ConnectorTypeId {
   POSTGRES = "postgres",
   MYSQL = "mysql",
   SQLSERVER = "sqlserver",
   MONGO = "mongodb",
+}
+
+export enum PropertyCategory {
+  PROPS_BASIC = "PROPS_BASIC",
+  PROPS_ADVANCED = "PROPS_ADVANCED",
+  FILTERS = "FILTERS",
 }
 
 /**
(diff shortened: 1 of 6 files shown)


ui/packages/ui/src/app/shared/Utils.ts
```diff
@@ -29,6 +36,74 @@ export function getConnectorTypeDescription(connType: ConnectorType): string {
...
+  const connProperties: ConnectorProperty[] = [];
+  for (const propDefn of propertyDefns) {
+    if (isInCategory(propDefn, category)) {
+      connProperties.push(propDefn);
+    }
+  }
+  return connProperties;
+}
+
+export function isInCategory(propertyDefn: ConnectorProperty, category: PropertyCategory) {
```
indraraj, line 71: Hoping that this logic moves to the backend. If not we could put this in a separate file of itself(as down the line more prop may also be added). WDYT?

indraraj (approved):
Looks good to me.

ui/packages/ui/src/app/shared/Utils.ts
mdrillin: @indraraj yes if we are ultimately forced to leave this on ui side, it should be separated out better.  My thinking is that this is just a temporary solution so that we can keep making progress on the components.  I'll just go ahead and merge this - maybe I will redo a little today if time permits

PR #46, opened 2020-08-28
Using property optional category, connecting test validation

mdrillin:
Made some adjustments to categorizing properties, and also made some changes to maintaining of the property values.  Did preliminary wiring of validation/save from the components.
- The assignment of property category was moved into CategoryUtil.  The optional category attribute is set on ConnectorProperty based on the property name.  Hopefully we can just remove this if the backend service can do the categorization.
- In CreateConnectorPage, now maintaining a map of the property values for each category separately.  These are then supplied to the appropriate wizard component for init of its forms.  Seemed easier to maintain this way, then upon finish merge the category maps to create the ConnectorConfiguration for submittal to the createConnector service.
- Removed the 'onSaveProperties' from the components for now.  We'll need to iron this out as UX design proceeds.
- Add temporary 'Validate' buttons in the FiltersStepComponent and ConfigureConnectorTypeComponent.  The validate actions send a couple properties to CreateConnectorPage which calls backend validate.  Temporarily added an alert there to show the validation result.
- An alert was also added upon click 'Finish', just to show the ConnectorConfig that would be submitted to the backend service.

### DIFF
--- a/ui/packages/services/src/connector/connector.service.ts
+++ b/ui/packages/services/src/connector/connector.service.ts
@@ -16,7 +16,7 @@
  */
 
 import {
-    ConnectionValidationResult, FilterValidationResult, PropertiesValidationResult
+    ConnectionValidationResult, FilterValidationResult, PropertiesValidationResult, ConnectorConfiguration
 } from "@debezium/ui-models";
 import {BaseService} from "../baseService";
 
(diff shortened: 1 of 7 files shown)


indraraj:
We could add the API fall back retry logic by using the fetch_retry wrapper method while calling API's.

indraraj (approved):
looks good addition until things get more clear.

mdrillin:
> We could add the API fall back retry logic by using the fetch_retry wrapper method while calling API's.

Yes, can be added on the next commit

PR #47, opened 2020-08-31
Added page heading in create connector

indraraj:
- Added the page heading to create connector page.

### DIFF
--- a/ui/packages/ui/src/app/Layout/Breadcrumb.tsx
+++ b/ui/packages/ui/src/app/Layout/Breadcrumb.tsx
@@ -37,7 +37,7 @@ import {
 	const count = items.length;
 	return (
 	  <AppBreadcrumb>
-		<Level gutter={'md'}>
+		<Level hasGutter={true}>
 		  <LevelItem>
 			<PFBreadcrumb>{items}</PFBreadcrumb>
 		  </LevelItem>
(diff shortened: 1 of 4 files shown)


uidoyen (commented):
@indraraj any particular reason to replace SASS with CSS?

indraraj:
> @indraraj any particular reason to replace SASS with CSS?

Honestly no such particular reason it's just that I have been working with CSS lately. BTW I am totally in for using SCSS it would be nice.

indraraj:
rebased to latest master, reverted to use the SCSS. 

@uidoyen @mdrillin I was thinking since we want to use SCSS we should then phase out the CSS and use SCSS uniformly across the app and configure the proper loader to handle the SCSS files. WDYT?

uidoyen:
@in

> rebased to latest master, reverted to use the SCSS.
> 
> @uidoyen @mdrillin I was thinking since we want to use SCSS we should then phase out the CSS and use SCSS uniformly across the app and configure the proper loader to handle the SCSS files. WDYT?

I think you are right. Let's merge this and then we can continue.

uidoyen (approved):

uidoyen:
> @in
> 
> > rebased to latest master, reverted to use the SCSS.
> > @uidoyen @mdrillin I was thinking since we want to use SCSS we should then phase out the CSS and use SCSS uniformly across the app and configure the proper loader to handle the SCSS files. WDYT?
> 
> I think you are right. Let's merge this and then we can continue.

PostCSS loader is already configured and SCSS is working fine.

indraraj:
> > @in
> > > rebased to latest master, reverted to use the SCSS.
> > > @uidoyen @mdrillin I was thinking since we want to use SCSS we should then phase out the CSS and use SCSS uniformly across the app and configure the proper loader to handle the SCSS files. WDYT?
> > 
> > 
> > I think you are right. Let's merge this and then we can continue.
> 
> PostCSS loader is already configured and SCSS is working fine.

it would be better if we could use .scss file also

PR #48, opened 2020-09-02
Adding data options form

uidoyen:
- Added data options form with basic setup.
- Snapshot Properties TBD


mdrillin (approved):
We will need to rework the 'DataOptionsForm' - it should dynamically create the form based on a supplied list of ConnectorProperties.  But this is a good start, merging.

PR #52, opened 2020-09-03
adding inital ui code

indraraj:
-  Added UI for Filters Step

-  Update the @patternfly version to use the TreeView component.

Will handle backend integration in subsequent pr.

### DIFF
--- a/ui/packages/ui/package.json
+++ b/ui/packages/ui/package.json
@@ -61,8 +61,8 @@
   "dependencies": {
     "@debezium/ui-models": "^1.0.0",
     "@debezium/ui-services": "^1.0.0",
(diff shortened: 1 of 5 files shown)


mdrillin (approved):
Looks like a good start.  For the switch, I'd personally like to see it 'on' by default with 'include' label and message under the filter entry.  Or better yet a toggle which shows both available states...  But what you have is according to the current design - we can discuss it with UX

indraraj:
> Looks like a good start. For the switch, I'd personally like to see it 'on' by default with 'include' label and message under the filter entry. Or better yet a toggle which shows both available states... But what you have is according to the current design - we can discuss it with UX

sure :+1:

PR #53, opened 2020-09-03
More changes for properties

mdrillin:
Some additional work to supply property defns correctly to the page components, and other misc fixes
- some changes/additions to the property categories in CategoryUtil.  Tried to further refine properties to their category/sub-group. Again this is a temporary solution which will be adjusted if this property metadata is supplied from the backend service.  Also setting 'required' on some properties in the same function so that it can be utilized.
- removed Edit button from ConnectorListItem, per feedback in the last meeting.
- removed connector temp data in ConnectorsPage - now showing empty state.  Made some adjustments for display of the create button and wording
- in Utils, added specific functions to get the property definitions according to category - so they can be supplied to the page components.  It is the responsibility of the page component to group them appropriately using the more specific category/subgroup info.
- renamed DbzLabelIcon to HelpInfoIcon, and removed a duplicate.


indraraj (approved):
looks good.

uidoyen (approved):

PR #62, opened 2020-09-09
Adding optional orderInCategory attribute to UI property model

mdrillin:
Adds an optional 'orderInCategory' attribute to the ConnectorProperty UI model, to assist in correct ordering of properties within a category/sub-group.

- An extra step was added in CategoryUtil which iterates the properties, assigning ordering to the properties within each category.
- The orderInCategory attribute is then utilized in ConfigureConnectorTypeForm to sort the property definitions

This is still very much a work-in-progress.  It is expected that the service layer will supply the category/group info - when that is ready, the CategoryUtil assigned categories will be removed and the ordering reworked.  In a similar way, we could also customize additional layout info for each property within the UI if needed (for example maybe row / column positioning or sizing).

### DIFF
--- a/ui/packages/models/src/ui.model.ts
+++ b/ui/packages/models/src/ui.model.ts
@@ -81,6 +81,7 @@ export interface ConnectorProperty {
     description: string;
     displayName: string;
     name: string;
+    orderInCategory?: number;
     required?: boolean;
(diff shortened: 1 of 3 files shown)


uidoyen (approved):
Looks good.

PR #63, opened 2020-09-10
Fixed the filter page issues

indraraj:
- Update the UI for Invalid filter, No matching content.

- Removed the switch and introduced the toggle for include/exclude.

- Uplifted the filter values to the conectConnectorPage and implement entered value persistence in moving next/back.

- Updated the bg color for disabled connector card.


mdrillin (approved):
looks good - I like the include/exclude toggle group.

PR #64, opened 2020-09-10
Adding Runtime options step to wizard

mdrillin:
Added a component for the final wizard step - Runtime Options.  The properties are grouped into 'Engine properties' and 'Heartbeat properties' per the current mockup.
- Reworked the categories for Data Options and Runtime Options to better match the design doc.
- added SSL properties and a few other missing properties.  SSL will not be displayed initially - it's just there for completeness
- Added the RuntimeOptionsComponent and RuntimeOptionsForm.  Validation still needs to be hooked up
Much of this will require more rework whenever the service is in place which provides additional property metadata (eg category, ordering, etc).


uidoyen (approved):
Looks good.

PR #69, opened 2020-09-11
Data Options Form based on properties

uidoyen:
- Added Select type component
- Added Checkbox component
- Added Switch component
- Added input type properties based on UI under Category Utils

Note: Added static options value to fill the drop-down for now, Will remove those with actually values


mdrillin (approved):
Good progress on the Data Options.  Will go ahead and merge - we can continue refining and finish it out from here.

PR #74, opened 2020-09-14
Re-positioning the properties section

uidoyen:
- Moved Snapshot properties section at the top
- Modified FormSelectComponent to pash allowedValues
- Replaced static drop down values with the actual one(allowedValues)


mdrillin (approved):
The layout on the data options page is improved.  There may be additional tweaks as we get more feedback, but this looks good to me.

PR #75, opened 2020-09-14
Refining properties and adapting to service

mdrillin:
Additional work to adapt to the service additions and refine properties 
- updated the ConnectorProperty model: category and isMandatory are no longer optional - they are supplied by the service.  Added 'allowedValues' attribute.  Aligned the category names to match those being supplied by the service.
- CategoryUtils - Removed specification of category and required since they are now supplied by the service.  Added some missing properties and fixed some ordering.
- ConfigureConnectorTypeForm - did some reorganization to put the properties into titled areas.  Also added some logic which hides the publication group if 'plugin' != Pgoutput.
- Created a generic 'FormComponent' - this component handles the logic to determine which specific form element is suitable for the supplied ConnectorProperty.  It will require further refinement - for example determining checkbox vs switch for booleans
- Now utilizing the FormComponent in ConfigureConnectorTypeForm

### DIFF
--- a/ui/packages/models/src/ui.model.ts
+++ b/ui/packages/models/src/ui.model.ts
@@ -68,22 +68,24 @@ export interface ConnectionValidationResult {
  */
 // tslint:disable-next-line: interface-name
 export interface ConnectorProperty {
-    category?: 'BASIC' | 
-    'ADVANCED_GENERAL' | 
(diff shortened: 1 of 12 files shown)


indraraj (approved):
Tested locally, looks good.

PR #83, opened 2020-09-15
Updated the property step and filter request payload

indraraj:
- Wired the property page forms

- Removed the hardcoded value for filter request.

### DIFF
--- a/ui/packages/ui/src/app/pages/createConnector/connectorSteps/ConfigureConnectorTypeForm.tsx
+++ b/ui/packages/ui/src/app/pages/createConnector/connectorSteps/ConfigureConnectorTypeForm.tsx
@@ -22,10 +22,7 @@ export interface IConfigureConnectorTypeFormProps {
   basicPropertyValues: Map<string, string>;
   advancedPropertyDefinitions: ConnectorProperty[];
   advancedPropertyValues: Map<string, string>;
-  onValidateProperties: (
-    connectorProperties: Map<string, string>,
-    category: PropertyCategory
-  ) => void;
+  onValidateProperties: (basicPropertyValues: Map<string,string>,advancePropertyValues:Map<string,string>) => void;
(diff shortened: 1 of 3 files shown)


ui/packages/ui/src/app/pages/createConnector/connectorSteps/ConfigureConnectorTypeForm.tsx
```diff
@@ -56,58 +78,85 @@ export const ConfigureConnectorTypeForm: React.FunctionComponent<IConfigureConne
...
+    _.union(
+      basicPropertyDefinitions,
+      advancedGeneralPropertyDefinitions,
+      advancedReplicationPropertyDefinitions,
+      advancedPublicationPropertyDefinitions
+    )
+  );
+
+  const handleSubmit = (valueMap: Map<string, string>) => {
+    // Validate the basic properties
```
mdrillin, line 133: I think the basic and advanced properties could be combined - would result in just a single call to the backend.  wdyt?

indraraj: I separated as I thought we might handle(logically) the "valid"/"invalid" response differently to basic and advance property.

mdrillin: I think all properties should be validated at once, but right, we may want to handle the response differently depending on whether the problem property is basic or advanced.

indraraj: yeah, we can check it with Gunnar & Team if they want the user to move ahead if only basic props are valid

mdrillin (approved):
Looks good

PR #84, opened 2020-09-15
rely on backend for property ordering

mdrillin:
Removed the optional orderInCategory attribute from ConnectionProperty.  Now relying on the order of properties within each category from the backend service


ui/packages/ui/src/app/pages/createConnector/connectorSteps/ConfigureConnectorTypeForm.tsx
```diff
@@ -34,14 +34,7 @@ export const ConfigureConnectorTypeForm: React.FunctionComponent<IConfigureConne
   const basicValidationSchema = {};
 
   const formatPropertyDefinitions = (propertyValues: ConnectorProperty[]) => {
-    const propertyValuesCopy = JSON.parse(JSON.stringify(propertyValues));
```
indraraj, line 37: Somehow this operation ends up modifying props(basicPropertyDefinitions in configureConnectorTypeForm) and state(in CreateConnectorPage) value direclty.  

i am adding "JSON.parse(JSON.stringify(propertyValues))" to break the reference else it may break the handling in other places.

indraraj (approved):
Updated one change, the rest looks good to me.

uidoyen (approved):

PR #89, opened 2020-09-16
Adjustments to form components

mdrillin:
Address a couple issues with form components
- allowedValues and defaultValue are optional in the model
- ConfigureConnectorTypeForm - the value is included for validation if a property has no default, or if the form value is different than the default
- other adjustments to form component for init


uidoyen (approved):
Looks good.

PR #96, opened 2020-09-17
fixing properties are not passing correct value to form

uidoyen:


mdrillin (approved):
fixes error observed with form properties

PR #97, opened 2020-09-17
Reorganize validation, hook up connector create

mdrillin:
Work to incorporate DataOptions and RuntimeOptions validation, hooked up connector create
- the DataOptionsComponent and RuntimeOptionsComponent were organized similar to the Connection properties step.
- added validateOptionProperties function to CreateConnectorPage - for use by the DataOptions and RuntimeOptions components
- some minor renaming in CreateConnectorPage
- hooked up the call to backend createConnector service upon finish.  It displays inline toast on failure and stays on the last wizard page.


indraraj (approved):
Tested locally, looks good.

PR #104, opened 2020-09-18
validation on data and runtime options page

indraraj:
- added the validation on the data and runtime option page.


mdrillin (approved):
Tested locally - nice progress

PR #109, opened 2020-09-18
reorganize data options, add connector name input

mdrillin:
- Reorganized Data Options step to better match the mockups
- Reorganized Runtime Options step - put property groups into accordions, to be consistent with other steps
- added connector name property into the basic properties.  The connection name is displayed at the top of the property configuration page.  Now passing the entered connector name upon create of the connector.
- other minor wording changes and clean up

### DIFF
--- a/ui/packages/ui/src/app/pages/createConnector/CreateConnectorPage.tsx
+++ b/ui/packages/ui/src/app/pages/createConnector/CreateConnectorPage.tsx
@@ -29,8 +29,11 @@ import {
   getBasicPropertyDefinitions,
   getDataOptionsPropertyDefinitions,
   getRuntimeOptionsPropertyDefinitions,
+  isDataOptions,
+  isRuntimeOptions,
   mapToObject,
-  PropertyCategory
+  PropertyCategory,
+  PropertyName,
 } from "src/app/shared";
 import {
   ConfigureConnectorTypeComponent,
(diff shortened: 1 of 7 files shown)


ui/packages/ui/src/app/pages/createConnector/CreateConnectorPage.tsx
```diff
@@ -207,10 +219,13 @@ export const CreateConnectorPage: React.FunctionComponent = () => {
   ): void => {
     setBasicPropValues(basicPropertyValues);
     setAdvancedPropValues(advancePropertyValues);
+    // Don't include connector name for validation
+    const basicForValidation = new Map<string,string>(basicPropertyValues);
+    basicForValidation.delete(PropertyName.CONNECTOR_NAME);
```
indraraj, line 224: why do we need to delete the CONNECTOR_NAME before calling validation?

indraraj (approved):
Tested locally looks good.

ui/packages/ui/src/app/pages/createConnector/CreateConnectorPage.tsx
indraraj: as it's currently already getting passed to validate in the filter step.

mdrillin: I removed it because it is an extra value that is not needed for backend validation.  However, it appears that the backend just ignores it - so it should not be a problem to leave it for now...

PR #110, opened 2020-09-21
Validation update and persistence of user value on 2nd step, proper e…

indraraj:
- Updated the validation on the connector property step(2nd). Now if the user updates the form after validation or by going back he is required to revalidate the form values.

- Persistence of the user entered values on the connector property step while going back and forth.

- Proper handling of the error in case of invalid filter, by identifying the problematic filter and showing msg coming from the backend. 
image


mdrillin (approved):
tested locally, looks good

PR #111, opened 2020-09-21
Send properties to backend

uidoyen:
- Improved display of errors coming from backend service when a form is validated.
- Fixed connector type cards have different heights
- Fixed step 4 and 5 not sending data to backend and moved Format Property Definitions function to Utils.

### DIFF
--- a/ui/packages/ui/src/app/pages/createConnector/CreateConnectorPage.tsx
+++ b/ui/packages/ui/src/app/pages/createConnector/CreateConnectorPage.tsx
@@ -2,7 +2,8 @@ import {
   ConnectionValidationResult,
   ConnectorProperty,
   ConnectorType,
-  PropertiesValidationResult
+
+  PropertiesValidationResult, PropertyValidationResult
 } from "@debezium/ui-models";
 import { Services } from "@debezium/ui-services";
 import {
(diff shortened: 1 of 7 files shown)


ui/packages/ui/src/app/pages/createConnector/connectorSteps/ConnectorTypeStepComponent.tsx
```diff
@@ -42,7 +42,7 @@ export const ConnectorTypeStepComponent: React.FunctionComponent<IConnectorTypeS
         : newId;
     props.onSelectionChange(newSelection);
   };
-
+  console.log(props.connectorTypesList)
```
mdrillin, line 45: please remove console.log or add ignore - build error

ui/packages/ui/src/app/pages/createConnector/CreateConnectorPage.tsx
```diff
@@ -337,6 +336,23 @@ export const CreateConnectorPage: React.FunctionComponent = () => {
     initPropertyValues();
   }, [connectorTypes]);
 
+  const ConnectionPropertiesError = ({ connectionPropsMsg }) => {
+    if(connectionPropsMsg.length !== 0){
+      return(
+        <ul>
+          {connectionPropsMsg.map((item, index) => (
+            <li key={index}>{item.property}: {item.message}</li>
```
mdrillin, line 344: The display name of the property with error should be shown, so that the user can identify the problem property on the form

mdrillin (approved):
Looks good, thanks for updating

PR #114, opened 2020-09-21
Adjustments for numeric in form

mdrillin:
- setting FormInputComponent type based upon the property type
- if property type INT or LONG and no defaultValue, setting initial value of 0


indraraj (approved):
looks good.

uidoyen (approved):

PR #117, opened 2020-09-22
Minimizing property values sent to backend for validation / creation

mdrillin:
Added Util function which minimizes the property values being set to the backend for validation / connector creation.  This essentially is now relying on the backend to use the appropriate default values for the properties.


uidoyen (approved):
Looks good!

PR #128, opened 2020-09-23
Improvements to form layouts

mdrillin:
Improved layout of properties by shortening some input components similar to mockup layouts
- added optional attribute 'gridWidth' to ConnectorProperty
- added function in Utils to set the gridWidth attribute on incoming properties
- wrapped the Accordion components to shorten overall form input widths to 75% of available space
- property gridWidth attributes are then used in the form grids to display appropriate width elements
- also improved the filter popover text and added a link to regex information

### DIFF
--- a/ui/packages/models/src/ui.model.ts
+++ b/ui/packages/models/src/ui.model.ts
@@ -86,6 +86,7 @@ export interface ConnectorProperty {
     defaultValue?: any;
     description: string;
     displayName: string;
+    gridWidth?: number;
     name: string;
     isMandatory: boolean;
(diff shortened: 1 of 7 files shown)


indraraj (approved):

indraraj:
@mdrillin wanted to check if the "Connector name" field should be mandatory?
and WDYT about using the "ExpandableSection" component instead of "Accordion" for maintaining the collapsible form sections?

mdrillin:
@indraraj imo the connector name should be mandatory because the connector create will fail without it.  
Yes nice catch - I think the 'ExpandableSection' would be more appropriate than 'Accordion' for the form sections - we can change those over

PR #129, opened 2020-09-24
added early finish option and disable filter next for invalid case

indraraj:
- Added option to skip and last 3 steps and early finish.

- Disable the Next and Finish button on the filter page for the invalid scenarios.

- Screenshots.
image
image

### DIFF
--- a/ui/packages/ui/src/app/components/FilterTreeComponent.tsx
+++ b/ui/packages/ui/src/app/components/FilterTreeComponent.tsx
@@ -6,7 +6,7 @@ import {
   Title,
   TreeView,
 } from "@patternfly/react-core";
-import { CubesIcon } from "@patternfly/react-icons";
+import { CubesIcon, ExclamationCircleIcon } from "@patternfly/react-icons";
 import React from "react";
 import { PageLoader } from ".";
 import { ApiError } from "../shared";
@@ -18,6 +18,7 @@ export interface IFilterTreeComponentProps {
   loading: boolean;
   apiError: boolean;
   errorMsg: Error;
+  invalidMsg: Map<string,string> | undefined;
 }
 export const FilterTreeComponent: React.FunctionComponent<IFilterTreeComponentProps> = (
   props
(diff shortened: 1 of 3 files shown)


uidoyen:
@indraraj I was thinking if you can show the final floating error message "Creation of the connector failed." at the same place where we are showing the API validation errors to maintain consistency. WDYT?

indraraj:
> @indraraj I was thinking if you can show the final floating error message "Creation of the connector failed." at the same place where we are showing the API validation errors to maintain consistency. WDYT?

my thought the process was that even if the creation of the connector failed used can still move to the next page if validation is true, so we may want to show him this. (As in the above screenshot even the connector creation is failed used can still move to the next page, which implied in the validation msg)

indraraj:
> > @indraraj I was thinking if you can show the final floating error message "Creation of the connector failed." at the same place where we are showing the API validation errors to maintain consistency. WDYT?
> 
> my thought the process was that even if the creation of the connector failed used can still move to the next page if validation is true, so we may want to show him this. (As in the above screenshot even the connector creation is failed used can still move to the next page, which implied in the validation msg)

also if you check on the last page since its the final step I am there removing the validation msg on connector creation fail.

uidoyen:
> > > @indraraj I was thinking if you can show the final floating error message "Creation of the connector failed." at the same place where we are showing the API validation errors to maintain consistency. WDYT?
> > 
> > 
> > my thought the process was that even if the creation of the connector failed used can still move to the next page if validation is true, so we may want to show him this. (As in the above screenshot even the connector creation is failed used can still move to the next page, which implied in the validation msg)
> 
> also if you check on the last page since its the final step I am there removing the validation msg on connector creation fail.

Screenshot from 2020-09-24 16-29-34

This is what I am getting in the last step.

indraraj:
> > > > @indraraj I was thinking if you can show the final floating error message "Creation of the connector failed." at the same place where we are showing the API validation errors to maintain consistency. WDYT?
> > > 
> > > 
> > > my thought the process was that even if the creation of the connector failed used can still move to the next page if validation is true, so we may want to show him this. (As in the above screenshot even the connector creation is failed used can still move to the next page, which implied in the validation msg)
> > 
> > 
> > also if you check on the last page since its the final step I am there removing the validation msg on connector creation fail.
> 
> Screenshot from 2020-09-24 16-29-34
> 
> This is what I am getting in the last step.

this is not the last step its "Data Option" page the last step will be "Runtime Option"

uidoyen:
> > > > > @indraraj I was thinking if you can show the final floating error message "Creation of the connector failed." at the same place where we are showing the API validation errors to maintain consistency. WDYT?
> > > > 
> > > > 
> > > > my thought the process was that even if the creation of the connector failed used can still move to the next page if validation is true, so we may want to show him this. (As in the above screenshot even the connector creation is failed used can still move to the next page, which implied in the validation msg)
> > > 
> > > 
> > > also if you check on the last page since its the final step I am there removing the validation msg on connector creation fail.
> > 
> > 
> > Screenshot from 2020-09-24 16-29-34
> > This is what I am getting in the last step.
> 
> this is not the last step its "Data Option" page the last step will be "Runtime Option"

Yeah, I was bit late to update the comment :smiley:

uidoyen (approved):
LGTM!

PR #130, opened 2020-09-24
Re-validating if boolean or dropdown selection value are changed

uidoyen:


indraraj (approved):
Tested locally looks good.

PR #139, opened 2020-09-24
Adapt to connector list response from service

mdrillin:
Some initial work to adapt the UI to the actual connector list from the backend service
- updated Connector model object in the UI
- created function in connector service to get the connector list
- adapted ConnectorListItem for display of the connector attributes
- created ConnectorStatus component for display of the connector status as a label with varying colors
- included the fetch in the ConnectorsPage, but it is still commented out until the backend code is merged.
- 'taskStates' just temp data for now - it will need to be hooked up.

### DIFF
--- a/ui/packages/models/src/ui.model.ts
+++ b/ui/packages/models/src/ui.model.ts
@@ -15,15 +15,15 @@
  * limitations under the License.
  */
 
-
 /**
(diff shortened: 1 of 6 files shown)


indraraj (approved):
looks good

PR #141, opened 2020-09-25
Moving the cluster dropdown into the header

uidoyen:
- Moving the cluster dropdown into the header
- Fixing wizard nav items not highlighting correctly


indraraj (approved):
Looks good.

indraraj:
It would be good if you also show the user name on the right side (any dummy value like "Developer").

uidoyen:
Yeah, I thought so but since we don’t have a specific reason to add right now. I skipped.  If you remember we removed action button earlier. We might discuss in upcoming meeting and add that back.

PR #143, opened 2020-09-28
added apply validation to filter step and folder restructure

indraraj:
- 3rd step: Disabled the "Next" and "Finish" to apply the filter expression changes before moving to the next step.

- Minor file/folder rearrangement.


uidoyen (approved):
tested locally. looks good

PR #145, opened 2020-09-28
Adding current and valid nav item color

uidoyen:
wizardSteps

Screenshot attached.


mdrillin (commented):
Tested locally.  I noticed that if you validate the last step, then go back - the last step does not indicate valid.

Also, we should verify with UX before implementing - they may have some input on whether this is consistent with patternfly best practices.  (emailed UX for input)

nding-anges:
Hi @mdrillin From the UX perspective, this is not consistent with Patternfly. Using a green color on these steps is not clear to indicate validate status. However, I am wondering if it has the necessity to show the validate status on the steps since there have already notifications to inform the user.

uidoyen:
As per feedback closing this PR

PR #146, opened 2020-09-28
Rework expansion areas in UI

mdrillin:
Replaced the accordion usages on the form pages with ExpansionSection - seems more appropriate according to patternfly examples.
- renamed ConfigureConnectorTypeComponent to ConfigureConnectorComponent
- replaced accordion usages with ExpansionSection
- indent form inputs to better distinguish the different sections.  Each area is titled, with exansion section below it.

example screenshots 
### Properties Step:
-----------------------------------------------------------------------------------------
ConnectionProperties

### Data Options Step:
-----------------------------------------------------------------------------------------
DataOptions

### Runtime Options Step:
-----------------------------------------------------------------------------------------
RuntimeOptions


uidoyen:
I think handling expansion with the title itself would be more generic then adding a separate expansion area. Expansion section looks different then actual UX. Maybe we can discuss with UX team.

uidoyen (commented):
I think handling expansion with the title itself would be more generic then adding a separate expansion area. Expansion section looks different then actual UX. Maybe we can discuss with UX team.

indraraj (approved):
Margin it now, improvement looks good. We can run the changes with UX later and can update if needed.

mdrillin:
@uidoyen agree, UX may have some input on this - we can adjust if needed after their review.  It may be better to just remove the titles and use the expansion text as the title.

PR #147, opened 2020-09-28
Add detail info to error alert

mdrillin:
Adds details message to the error alert - if there is a problem creating the connector.  If no message, then a generic message is added.

ErrorDetail1

ErrorDetails2


indraraj (approved):
looks good.

PR #148, opened 2020-09-29
Remove the error toast msg after 10 sec

indraraj:
- Remove the error Toast msg automatically after 10 sec.

- Moved the Toast msg in a different component.


mdrillin (approved):
Looks good, will merge.  I think it would be good to leave a close button on it too - then the user can dismiss earlier if they want.  wdyt?

indraraj:
> Looks good, will merge. I think it would be good to leave a close button on it too - then the user can dismiss earlier if they want. wdyt?

Yeah, i think it will good to have that. I will get that

PR #149, opened 2020-09-29
Fixing expansion panel

uidoyen:
Removing extra title for expansion area in Properties, Data Options and Runtime Options steps


mdrillin (approved):
I agree - this looks much better, thanks!

PR #152, opened 2020-09-29
Connector list page work for fetch of connectors list

mdrillin:
Additional work on the connector list page and connector items
- modify kafka connect cluster component to pass the clusterId on change.  This allows the ConnectorsPage to get the id directly from the context instead of fetching the connectors again to determine the id.
- remove call from ConnectorsPage to get clusters - no longer needed
- fix the fetch of the connectors list
- passing actual task state to ConnectorListItem (was hardcoded)
- adapt ConnectorListItem to recognize alternate PostgreSQL type name.


uidoyen (approved):
Tested locally. Looks good.

PR #170, opened 2020-10-05
Making backend base URL configurable

### LINKED ISSUE #168: JS/FE: Make backend base URL relative to the src URI or configurable
When the React app is loading deployed on a server it tries to query `localhost:8080` instead if the source domian where the JS was loaded from. Either load from the JS files source domain (just relative URLs) or make the base URL configurable.

FYI @gunnarmorling @mdrillin @indraraj @uidoyen

uidoyen:
Create/Generate `config.js` inside `build` or `dist` folder with the following code snippet and change `url `accordingly.

```
window.DEFAULT_CONFIG = {
  artifacts: {
      type: "rest",
      url: "http://localhost:8080/api/"
  },
  features: {
      readOnly: false
  },
  mode: "dev",
  ui: {
      contextPath: null,
      url: "http://localhost:8888/"
  }
};
```

closes #168

### DIFF
--- a/ui/packages/ui/src/index.html
+++ b/ui/packages/ui/src/index.html
@@ -8,13 +8,11 @@
 
   <meta id="appName" name="application-name" content="Debezium UI">
   <link rel="icon" type="image/x-icon" href="favicon.ico">
-
+  <script src="config.js"></script>
 </head>
 
 <body>
   <noscript>Enabling JavaScript is required to run this app.</noscript>
   <div id="root"></div>
-  <script src="config.js"></script>
 </body>
-
 </html>
--- a/ui/packages/ui/src/index.html
+++ b/ui/packages/ui/src/index.html
@@ -14,6 +14,7 @@
 <body>
   <noscript>Enabling JavaScript is required to run this app.</noscript>
   <div id="root"></div>
+  <script src="config.js"></script>
 </body>
 
 </html>
--- a/backend/src/main/java/io/debezium/configserver/rest/FrontendConfigResource.java
+++ b/backend/src/main/java/io/debezium/configserver/rest/FrontendConfigResource.java
@@ -1,0 +1,30 @@
+package io.debezium.configserver.rest;
+
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import java.net.URI;
+
+@Path("")
+public class FrontendConfigResource {
+
+    public static final String PROPERTY_BASE_URI = "ui.base.uri";
+    public static final String FRONTEND_CONFIG_ENDPOINT = "/config.js";
+    public static final String DEFAULT_BASE_URI = "http://localhost:8080";
+
+    @ConfigProperty(name = PROPERTY_BASE_URI, defaultValue = DEFAULT_BASE_URI)
+    URI UiBaseURI;
+
+    @Path(FRONTEND_CONFIG_ENDPOINT)
+    @GET
+    @Produces("application/javascript")
+    public String getFrontendConfig() {
+        return "window.DEFAULT_CONFIG={" +
+                "artifacts:{"
+                + "type:\"rest\",url:\"" + UiBaseURI + ConnectorResource.API_PREFIX + "/\"" +
+                "},features:{readOnly:false},mode:\"prod\"};";
+    }
+
+}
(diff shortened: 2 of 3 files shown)


indraraj (approved):
Looks good to me if this solves the issue.

ui/packages/ui/src/index.html
```diff
@@ -14,6 +14,7 @@
 <body>
   <noscript>Enabling JavaScript is required to run this app.</noscript>
   <div id="root"></div>
+  <script src="config.js"></script>
```
rk3rn3r, line 17: 1. Should I make this an endpoint in the Backend that returns the right config? What happens if the file can't be loaded?

2. Don't we have a timing issue here when someone starts using the React app before that config.js is returned? Maybe move this to the `<head>` to avoid such timing issues?

uidoyen: 1. Yes, It needs to generate ‘config.js’ with required configuration. If file doesn’t exist or can’t be loaded. It will take default configuration which is under `ui/packages/services/src/config/config.service.ts`
2.  App bundle loads after ‘config.js’ so it shouldn’t be an issue.

gunnarmorling:
So how would the value within _config.js_ be determined then? Is the expectation that the backend somehow dynamically generates this file?

Tbh. this all seems over-complicated, there surely must be a simple way for a React app to obtain its base URL, no? E.g. how about [this answer](https://stackoverflow.com/questions/48134785/how-to-set-a-base-url-for-react-router-at-the-app-level) on SO?

ui/packages/ui/src/index.html
rk3rn3r: The 404 is an issue. People running the container and monitoring it would not like to see 404 in their logs. I fixed that by having the backend always return config. Is there any way we can remove the `mode` property and the other stuff from the response?
I will log an issue for cleaning up that feature and also fix the possible timing issue, because:

(2): No, that's obviously not a problem, but the other way around. The app is already loaded but `config.js` did not return, then it will use the default config.

rk3rn3r:
@gunnarmorling it's good for now. It should be working with the changes I applied now, but it's ugly.
There are scenarios especially in the enterprise sector that makes it necessary to have config for that. e.g. like you run one instance to serve the assets (JS files, CSS files, webfonts etc) in a frontend network and run a separate second instance in a backbone network to serve the `/api` endpoints. The solution we have now will solve the current deployment issue but we need to invest more time later, when this really will be used.

rk3rn3r:
@uidoyen  @indraraj 

what's the minimal fields I need to return for prod? and can I return the same for dev?

```
window.DEFAULT_CONFIG = {
  artifacts: {
      type: "rest",
      url: "http://localhost:8080/api/"
  },
  features: {
      readOnly: false
  },
  mode: "dev",
  ui: {
      contextPath: null,
      url: "http://localhost:8888/"
  }
};
```

uidoyen:
> features

Possible config for `prod`
```
window.DEFAULT_CONFIG = {
  artifacts: {
      type: "rest",
      url: "http://localhost:8080/api/"
  },
  features: {
      readOnly: false
  },
  mode: "prod",
  ui: {
      contextPath: null,
      url: "http://localhost:8888/"
  }
};
```

> features

Possible config for `dev`
```
window.DEFAULT_CONFIG = {
  artifacts: {
      type: "rest",
      url: "http://localhost:8080/api/"
  },
  features: {
      readOnly: false
  },
  mode: "dev",
  ui: {
      contextPath: null,
      url: "http://localhost:8888/"
  }
};
```

uidoyen:
Possible config for `prod`
```
window.DEFAULT_CONFIG = {
  artifacts: {
      type: "rest",
      url: "http://localhost:8080/api/"
  },
  features: {
      readOnly: false
  },
  mode: "prod",
  ui: {
      contextPath: null,
      url: "http://localhost:8888/"
  }
};
```

Possible config for `dev`
```
window.DEFAULT_CONFIG = {
  artifacts: {
      type: "rest",
      url: "http://localhost:8080/api/"
  },
  features: {
      readOnly: false
  },
  mode: "dev",
  ui: {
      contextPath: null,
      url: "http://localhost:8888/"
  }
};
```

gunnarmorling:
Ok, so I still l don't understand why we have this 40+ LOC monstrosity instead of simply evaluating `window.location.host`, like it's done in JS since the dawn of time ;) But if what we got here works, let's move forward with it now and keep it for later to hopefully simplify this.

indraraj (approved):

rk3rn3r:
It's 33 lines now

PR #173, opened 2020-10-06
align connector type card text, add connector name check

mdrillin:
- Set card headers to same height, to better align card text
- Add check of connector name to avoid duplicate - checks entry against already existing connector names


ui/packages/ui/src/app/pages/createConnector/CreateConnectorPage.tsx
```diff
@@ -307,11 +309,31 @@ export const CreateConnectorPage: React.FunctionComponent = () => {
     validateOptionProperties(propertyValues, propertyCategory);
   };
 
+  const validateConnectionName = (connName: string): string => {
+    const currentNames = location.state?.connectorNames;
```
indraraj, line 313: Hey, @mdrillin I was thinking if we need to add ignore case to name. Anyway, I have asked the same to gunner and Rene on the GitHub issue.

indraraj (approved):
Tested locally, looks good.

PR #185, opened 2020-10-06
Fix a few UI issues

mdrillin:
- address brief flashing of alert by adding validationInProgress state.  If validation is in progress, a spinner is shown
- address similar issue on filters page - a spinner is shown on the lower panel when loading
- address problem with deselect of connector type.  Not resetting properties if connectorType is deselected.  Also disabling next button if no type is selected
- click finish on steps 2-5 will now go to the final review step instead of create connector.  Create only happens from the review step now.

### DIFF
--- a/ui/packages/ui/src/app/pages/createConnector/CreateConnectorPage.tsx
+++ b/ui/packages/ui/src/app/pages/createConnector/CreateConnectorPage.tsx
@@ -15,6 +15,7 @@ import {
   LevelItem,
   PageSection,
   PageSectionVariants,
+  Spinner,
   TextContent,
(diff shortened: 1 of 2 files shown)


indraraj (approved):
Tested locally, changes look good. The loader thing is really great addition.

indraraj:
@mdrillin  while testing I find some inconsistency in how wizard steps(Left side) are enabled in performing early finish. we can look at this in subsequent Pr's merging this one.

indraraj:
one more issue on the table selection page "Next" button is not getting disabled after changing the filter expression

PR #186, opened 2020-10-07
Masking password on the Review page

uidoyen:


ui/packages/ui/src/app/shared/Utils.ts
```diff
@@ -340,6 +340,15 @@ export function mapToObject(inputMap: Map<string, string>) {
   return obj;
 }
 
+export function maskPropertyValues(inputObj:{}){
+  for (const key in inputObj) {
+  	if(key === "database.password") {
```
indraraj, line 345: It might be nitpicking but just one thing it would be great if we match the masked length to password length. Do let me know if you think its an overkill.

indraraj (approved):
Looks good.

PR #190, opened 2020-10-08
Fixing undefined Connector Names from location state 

uidoyen:
Redirecting create connector page to `/app` if connector names from location state is undefined


mdrillin (approved):
tested locally, looks good

PR #194, opened 2020-10-12
Added the polling to create connector page and hooked up delete button

indraraj:
- Hooked up the `/api/connectors/{cluster}/{connector-name}` to delete button on the connector list.

- Added the polling to `/api/connectors/{cluster}` to get the updated list of connector at the interval of 10 sec


uidoyen (approved):
Tested locally, looks good.

PR #196, opened 2020-10-12
Hash route

indraraj:
added the hash`#` in route for bypassing the unknown URI issue altogether.

### DIFF
--- a/ui/packages/ui/src/app/pages/connectors/ConnectorsPage.tsx
+++ b/ui/packages/ui/src/app/pages/connectors/ConnectorsPage.tsx
@@ -47,7 +47,7 @@ export const ConnectorsPage: React.FunctionComponent = (props) => {
       return conn.name;
     });
     history.push({
-      pathname: "/app/create-connector",
+      pathname: "/create-connector",
       state: { value: appLayoutContext.clusterId, connectorNames },
     });
   };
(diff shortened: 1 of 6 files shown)


ui/packages/ui/src/app/pages/connectors/ConnectorsPage.tsx
```diff
@@ -47,7 +47,7 @@ export const ConnectorsPage: React.FunctionComponent = (props) => {
       return conn.name;
     });
     history.push({
-      pathname: "/app/create-connector",
+      pathname: "/create-connector",
```
rk3rn3r, line 50: maybe `/#create-connector`?

ui/packages/ui/src/app/shared/routes.tsx
```diff
@@ -5,18 +5,18 @@ export const ROUTES = [
...
       {
-        path: "/app",
+        path: "/",
         key: "APP_ROOT",
         exact: true,
         component: ConnectorsPage,
       },
       {
-        path: "/app/create-connector",
+        path: "/create-connector",
```
rk3rn3r, line 19: same here?

ui/packages/ui/src/app/pages/connectors/ConnectorsPage.tsx
indraraj: yes, it will be appended to the basename. so it will be like `#app//create-connector`

uidoyen (approved):
Tested locally. looks good.

PR #206, opened 2020-10-15
Minor fixes in Create Connector Wizard

mdrillin:
A few minor fixes from UX feedback
- fix case on wizard steps and breadcrumbs
- change to inline alert style and move above wizard buttons
- added 'Optional' label on wizard 'Additional properties' step (left commented out pending UX input)


indraraj (approved):
Tested locally, looks good.

PR #209, opened 2020-12-02
Pete's pr for showing use of webpack 5

indraraj:
Comments in original PR

> This is a bit out of date so there are conflicts.
The goal here is to show how to use webpack 5 and then also to expose components over it.
As part of that I had to break up components and make functionality like navigation external to the view.

- Reverted the changes in `docker-compose.yml ` file to run the backend in docker itsef. As per changes by @rk3rn3r 

- Resolved the conflicts in `CreateConnectorPage` and `ConnectorsPage`  files.

@pmuir @redallen


pmuir (approved):

mdrillin (approved):
tested locally, looks good - merging

PR #210, opened 2020-12-03
Updating the validate & next button position as per UX feedback

indraraj:
- Updated the validate & next button position as per UX feedback


mdrillin (changes requested):
The positioning of the buttons looks fine.   However, the enablement of 'Skip to finish' I think needs to be addressed.  Skip to finish should be disabled if the current page is incomplete or invalid. (I think it can use the same enablement logic as the 'Next' button).

indraraj:
> The positioning of the buttons looks fine. However, the enablement of 'Skip to finish' I think needs to be addressed. Skip to finish should be disabled if the current page is incomplete or invalid. (I think it can use the same enablement logic as the 'Next' button).

Thanks for pointing, I missed the regression effect on the skip button.

indraraj:
Reverted the `Skip to finish` button as per the current implementation.

mdrillin (approved):
looks good, thanks

PR #217, opened 2020-12-07
Adding skip to finish info box

uidoyen:


indraraj:
Hey @uidoyen I think as mention by Mark https://github.com/debezium/debezium-ui-poc/issues/205 we need to evaluate if this needs to be implemented as it is suggested by the UX team. 

@mdrillin WDYT?

mdrillin:
@uidoyen @indraraj - I reached out to UX via email to confirm the preferred approach.

mdrillin (approved):
Verified with UX that they prefer the info box approach.  Tested locally, looks good.  I will probably open a follow-up issue with a couple other minor changes.

Issue #220, opened 2020-12-07
Utilize toggle switch component in UI


mdrillin:
UX request that we use a toggle switch component for a couple of the boolean connector properties - 'TCP Keep Alive' and 'Drop slot on stop'

![Toggle](https://user-images.githubusercontent.com/1820403/101384349-ef328680-387f-11eb-8868-7714e6d6c91d.png)

mdrillin:
Addressed in pending PR #222

indraraj:
Merged.

PR #222, opened 2020-12-07
Utilizing switch component for some properties

mdrillin:
Addresses issue #220 - UX request to use toggle switch for a couple properties
- add 'BOOLEAN-SWITCH' to ui model to distinguish toggle switch props
- update FormComponent to check for BOOLEAN-SWITCH type
- update Utils.getFormattedProperties to handle the switch properties


indraraj (approved):
Tested locally looks good, I believe this https://github.com/debezium/debezium-ui-poc/issues/175 issue needs to be addressed separately.

PR #231, opened 2020-12-14
Fixing form fields value not being retained

uidoyen:


mdrillin (approved):
Tested the components - looks good!

Issue #126, opened 2020-09-23
UI: Create a duration component for forms


mdrillin:
Some properties require an entry of duration or units of time.  Create a component where user can optionally select the units for the input (e.g.  ms vs sec).  An screenshot is below

![DurationComponent](https://user-images.githubusercontent.com/1820403/94026451-4f388700-fd7f-11ea-94b1-efab3b54dc6c.png)

Example code in syndesis -
https://github.com/syndesisio/syndesis/blob/master/app/ui-react/packages/auto-form/src/widgets/FormDurationComponent.tsx

mdrillin:
Pending PR - #232

mdrillin:
PR merged - closing

PR #232, opened 2020-12-14
Adding duration component

mdrillin:
Adding a duration component, Issue #126 - and a couple other component improvements.  Some of the properties which utilize the duration component will need to have their titles and descriptions adjusted - I will log a follow-up request for that
- added FormDurationComponent, which allows the user to enter duration values in different units.  Regardless of the selected units, the value is always stored as milliseconds and provided as ms to the backend.
- added a new ConnectorProperty type 'DURATION' to the model and now designating the duration properties in Utils.getFormattedProperties.

Screenshot example of duration component:
durationComponent

- Added HelpInfoIcon to the checkbox and switch components, for display of the property descriptions, similar to other properties
- added new ConnectorProperty types of 'NON-NEG-INT' and 'NON-NEG-LONG' to the model.  This allows the FormInputComponent to disallow negative numbers for those designated properties.  Now designating some of the int properties as such in Utils.getFormattedProperties.  We can now disallow negative values up-front, instead of needing backend validation to see a failure.

### DIFF
--- a/ui/packages/models/src/ui.model.ts
+++ b/ui/packages/models/src/ui.model.ts
@@ -89,7 +89,19 @@ export interface ConnectorProperty {
     gridWidth?: number;
     name: string;
     isMandatory: boolean;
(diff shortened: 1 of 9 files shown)


indraraj (approved):
Tested Locally, great work @mdrillin

PR #233, opened 2020-12-15
Form component for mask and truncate feild

indraraj:
- Added the FormFragmentComponent to handle the `column.truncate.to.(d+).chars` &  `column.mask.with.(d+).chars` fields.

- Using the value entered in `n:` field to update the key and value in `column:` to update the corresponding value of the key.

- If the value entered in only any one of the feild then the combined value is ignored.

### DIFF
--- a/ui/packages/ui/src/app/components/formHelpers/FormFragmentComponent.tsx
+++ b/ui/packages/ui/src/app/components/formHelpers/FormFragmentComponent.tsx
@@ -1,0 +1,111 @@
+import {
+  Flex,
+  FlexItem,
+  FormGroup,
+  Grid,
+  GridItem,
+  InputGroup,
+  TextInput,
+} from "@patternfly/react-core";
+import { ExclamationCircleIcon } from "@patternfly/react-icons";
+import { useField } from "formik";
+import * as React from "react";
+import "./FormFragmentComponent.css";
+import { HelpInfoIcon } from "./HelpInfoIcon";
+
+export interface IFormFragmentComponentProps {
+  label: string;
+  description: string;
+  name: string;
+  fieldId: string;
+  helperTextInvalid?: any;
+  isRequired: boolean;
+  validated?: "default" | "success" | "warning" | "error" | undefined;
+  propertyChange: (name: string, selection: any) => void;
+  setFieldValue: (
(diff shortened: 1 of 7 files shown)


ui/packages/ui/src/app/components/formHelpers/FormFragmentComponent.tsx
```diff
@@ -0,0 +1,111 @@
...
+      }
+      helperTextInvalid={props.helperTextInvalid}
+      helperTextInvalidIcon={<ExclamationCircleIcon />}
+      fieldId={id}
+      validated={props.validated}
+    >
+      <InputGroup>
+        <Grid>
+          <GridItem span={9}>
+            <Flex className={'form-fragment-component-checkbox'}>
```
mdrillin, line 78: Remove 'checkbox' from this className

ui/packages/ui/src/app/components/formHelpers/FormFragmentComponent.tsx
```diff
@@ -0,0 +1,111 @@
...
+                  defaultValue={field.value.split("&&")[0]}
+                />
+              </FlexItem>
+            </Flex>
+          </GridItem>
+          <GridItem span={3}>
+            <Flex>
+              <FlexItem>n:</FlexItem>
+              <FlexItem>
+                <TextInput
```
mdrillin, line 96: Since this needs to be a positive int, we could probably set "min={"1"}" on the TextInput.  Could possibly also use 'onKeyPress' as in FormDurationComponent, to prevent entry of '-' and '.'

mdrillin (approved):
I think this looks good, just a couple minor suggestions noted, plus 
- for the column list input - user can leave a trailing comma which stays with the value (maybe that should be trimmed off - not sure if backend is ok with it?)  
- I was trying to think of a more descriptive component name instead of FormFragmentComponent -  maybe FormMaskOrTruncateComponent ?

indraraj:
Incorporated the suggestions.

- I will confirm once with Rene regarding min={'1'} and trailing comma issue.

- For the component name, I was trying to use a general name, updated the name to `FormMaskOrTruncateComponent` as of now.

PR #234, opened 2020-12-15
Fix duration titles, rework table tooltips

mdrillin:
- remove units from duration form component titles.  The user can select different units, so the titles were confusing
- remove tooltips from ConnectorStatus in the table - the tooltip didn't provide any additional info
- in ConnectorTask, changed tooltip to popover.  The popover icon will only show up if there are additional errors to display.
- added aria-label to the HelpInfoIcon.  Should help with accessibility for the help popovers


indraraj (approved):
Tested Locally, looks good.

PR #238, opened 2020-12-18
Adding toolbar to connector table

uidoyen:


ui/packages/ui/src/app/pages/connectors/ConnectorsTableComponent.tsx
```diff
@@ -129,8 +138,12 @@ export const ConnectorsTableComponent: React.FunctionComponent<IConnectorsTableC
   }, [alerts]);
 
   React.useEffect(() => {
-    const getConnectorsInterval = setInterval(() => getConnectorsList(), 10000);
-    return () => clearInterval(getConnectorsInterval);;
+    getConnectorsList();
+    const getConnectorsInterval = setInterval(() => {
+      getConnectorsList()
+      setCurrentCategory('Name') // reset table sorting to default
```
mdrillin, line 144: The intent of this useEffect is to refresh the table rows from the backend at periodic intervals.  When the table refreshes, it should not revert to the default sort.  Also, I think the interval should be shorter than 50sec - I would leave it at 10sec

mdrillin (changes requested):
It looks like the refresh needs more work.  I tested a bit - it looks like the 'currentCategory' state keeps reverting back to 'Name' when the table updates.

mdrillin:
Thanks, I'll merge this.  We can change the sort direction icon to a toggle button in a separate PR

PR #245, opened 2021-01-01
Added the remaining i18n work

indraraj:
Added remaining translation work.


indraraj:
tested locally, merging this.

PR #249, opened 2021-01-05
Mask hash salt component

indraraj:
Comments form @mdrillin original pr

> Adds UI component to handle the hash-salt masking property.
> 
> using same strategy for the value as the column truncation and masking component, but 3 parts are combined instead of 2. Different delimiters are used to separate the different segments
> added 'COLUMN_MASK_HASH_SALT' type in the model, so that the property can be distinguished from others
> made a couple of minor naming changes


mdrillin (approved):
tested locally - looks good!

PR #254, opened 2021-01-06
Allowing multiple mask/truncate prop defns

mdrillin:
Allow definition of multiple column mask/truncate properties.  WIP


ui/packages/ui/src/app/shared/Utils.ts
```diff
@@ -274,11 +274,15 @@ export function minimizePropertyValues (propertyValues: Map<string, string>, pro
...
-        // The value is of form : Cols&&n - where && is used as a separator
-        const [a,b] = value.split('&&');
-        const updatedKey = key.replace('(d+)',b);
-        if(a !== '' && b !== ''){
-          minimizedValues.set(updatedKey, a);
+        // The value value can have multiple entries (entries are separated by '@^')
+        // Example colsString&&number @^ colsString&&number
+        const entries = value.split("@^");
+        for(const entry of entries) {
+          const [a,b] = entry.split('&&');
```
indraraj, line 281: need to add a check for `entry !==""` cause in such case b will be `undefined`

ui/packages/ui/src/app/components/formHelpers/MaskOrTruncateItem.tsx
```diff
@@ -0,0 +1,99 @@
...
+  canDelete: boolean;
+  maskTruncateItemChanged: (rowId: number, maskTruncateValue: string) => void;
+  deleteMaskTruncateItem: (rowId: number) => void;
+}
+
+export const MaskOrTruncateItem: React.FunctionComponent<IMaskOrTruncateItemProps> = (
+  props
+) => {
+
+  const [currentColumns, setCurrentColumns] = React.useState<string | undefined>(props.columnsValue);
```
indraraj, line 25: for the React components initialized state from `props`, we should add the corresponding `useEffect` code to update the state value in case the props passed value is being changed.

PR #255, opened 2021-01-07
Added the wizard exit confirmation dialog on the browser back button

indraraj:
Added Pr for https://github.com/debezium/debezium-ui-poc/issues/67 Github issue

- Added the Wizard Exit confirmation dialog on the browser back button so that the user is warned of losing all the entered values before moving to the connector list page.


mdrillin (approved):
There is a lint issue with the imports, but I'll go ahead and merge.  Will follow up immediately with another fix PR and axios upgrade.  The back button works as expected now.

PR #260, opened 2021-01-08
Fixed the wizard back button issue

indraraj:
Fixed the issue https://github.com/debezium/debezium-ui-poc/issues/257

- Updated the wizard navigation call function form `goToStepByName` to `goToStepById`
_(after adding the required red asterisk mark work on steps name, some of the wizard names were no longer the string values)_


mdrillin (approved):
looks good!

PR #262, opened 2021-01-08
Adds multiple hash-salt property capability

mdrillin:
This adds the multi-row capability to the 'MaskHashSalt' component.  Took a similar approach as with the other multi-row property, breaking the item out into a separate component.

### DIFF
--- a/ui/packages/ui/src/app/shared/Utils.ts
+++ b/ui/packages/ui/src/app/shared/Utils.ts
@@ -287,13 +287,19 @@ export function minimizePropertyValues (propertyValues: Map<string, string>, pro
           }
         }
       }else if(propDefn.name.startsWith("column_mask_hash") && value !== "") {
-        // The value is of form : Cols&&Hash||Salt - where && and || are used as separators
-        const [cols, trailing] = value.split("&&");
-        if (trailing) {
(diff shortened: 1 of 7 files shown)


ui/packages/ui/src/app/shared/Utils.ts
```diff
@@ -287,13 +287,19 @@ export function minimizePropertyValues (propertyValues: Map<string, string>, pro
...
-        if (trailing) {
-          const [hash, salt] = trailing.split("||");
-          if(cols !== "" && hash !== "" && salt && salt !== ""){
-            const updatedKey = "column.mask.hash.([^.]+).with.salt.(.+)".replace("([^.]+)",hash).replace("(.+)",salt);
-            minimizedValues.set(updatedKey, cols);
+        // The value value can have multiple entries (entries are separated by '#$')
+        // Example Cols&&Hash||Salt #$ Cols&&Hash||Salt
+        const entries = value.split("#$");
+        for(const entry of entries) {
+          if (entry && entry !== "") {
```
indraraj, line 294: `""` itself is a false value `if(entry)` will also work as same

indraraj (approved):
Tested locally, looks good.

PR #265, opened 2021-01-12
Fixed the alert spinner in table selection step

indraraj:
Fix for removing the spinner in case of API error in the table selection step.


mdrillin (approved):
looks good

PR #272, opened 2021-01-13
Filter page

indraraj:
Fixed the issue https://github.com/debezium/debezium-ui-poc/issues/268

- Updated the "Table selection" step name to a more general "Filter definition".
- Made the "Filter definition" dynamic where the key values are being passed accordingly based on the Connector Type.
- Update the Label text to pick based on the Connector type, i.e form "schema/table" to "database/collection".


mdrillin (approved):
Tested locally - looks good!

Issue #271, opened 2021-01-12
Wizard next button enablement issue when switching connector types


mdrillin:
After introducing the mongo connector type, I've noticed an issue with the next button enablement in the wizard for the following scenario.

1) Select Mongo connector type, click next to go to properties
2) enter mongo properties and validate.  click next to go to filters page.
3) now in the wizard LH nav menu, click to go back to the first page (Connector type)
4) Choose the Postgres connector type and click next.

On the properties page, notice that the properties are empty.  However the Next button is enabled.   Clicking next results in an error.
The next button should be disabled until the properties are entered and validated

mdrillin:
Pending PR #273

indraraj:
Pr merged.

PR #273, opened 2021-01-14
Fix issue with wizard next button

mdrillin:
Fix for issue #271 - next button was remaining enabled after connector type change
- Reset some states after connectorType change, to re-init
- combining 2 ConfirmationDialog usages into 1
- Utils fix to ensure propNames are properly matched (was losing formatting after form validations due to formik name adjustments)


uidoyen (approved):
Tested locally. Looks good.

PR #274, opened 2021-01-15
Updated the browser back button confirm handling by using prompt

indraraj:
Fixed issue https://github.com/debezium/debezium-ui-poc/issues/258
Updated the browser back button confirm implementation code, now using `<Prompt>` available in react-router to check the back event and rendering custom UI.

### DIFF
--- a/ui/packages/ui/src/app/pages/createConnector/CreateConnectorComponent.tsx
+++ b/ui/packages/ui/src/app/pages/createConnector/CreateConnectorComponent.tsx
@@ -23,7 +23,7 @@ import {
 import _ from "lodash";
 import React, { Dispatch, ReactNode, SetStateAction } from "react";
 import { useTranslation } from "react-i18next";
-import { useHistory } from "react-router-dom";
+import { Prompt, useHistory } from "react-router-dom";
 import { ToastAlertComponent } from "src/app/components";
 import {
   ConfirmationButtonStyle,
@@ -147,7 +147,6 @@ export const CreateConnectorComponent: React.FunctionComponent<ICreateConnectorC
     runtimeOptionsPropValues,
     setRuntimeOptionsPropValues,
   ] = React.useState<Map<string, string>>(new Map<string, string>());
-  const [backEnable, setBackEnable] = React.useState(false);
 
   const [validateInProgress, setValidateInProgress] = React.useState(false);
   const [loading, setLoading] = React.useState(true);
@@ -157,11 +156,6 @@ export const CreateConnectorComponent: React.FunctionComponent<ICreateConnectorC
   const [
     showCancelConfirmationDialog,
     setShowCancelConfirmationDialog,
-  ] = React.useState(false);
-
-  const [
-    showBackConfirmationDialog,
-    setShowBackConfirmationDialog,
   ] = React.useState(false);
 
   const [connectionStepsValid, setConnectionStepsValid] = React.useState<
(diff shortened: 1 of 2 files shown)


mdrillin:
On click cancel in the wizard, the "Leave" button is now requiring 2 clicks to accept.  The first click is ignored.

ui/packages/ui/src/app/pages/createConnector/CreateConnectorComponent.tsx
```diff
@@ -1078,22 +1043,36 @@ export const CreateConnectorComponent: React.FunctionComponent<ICreateConnectorC
 
   return (
     <>
+      <Prompt
+        message={(location: any, action: any) => {
+          if (action === "POP") {
```
mdrillin, line 1048: if block can be removed

ui/packages/ui/src/app/pages/createConnector/CreateConnectorComponent.tsx
```diff
@@ -23,7 +23,7 @@ import {
 import _ from "lodash";
 import React, { Dispatch, ReactNode, SetStateAction } from "react";
 import { useTranslation } from "react-i18next";
-import { useHistory } from "react-router-dom";
+import { Prompt, useHistory } from "react-router-dom";
```
mdrillin, line 26: useHistory can be removed - not used

indraraj:
> On click cancel in the wizard, the "Leave" button is now requiring 2 clicks to accept. The first click is ignored.

Hey @mdrillin just tested again its seems to be going back with just one click on the "leave" button. let me know if you are getting this in any specific case.

mdrillin:
@indraraj - when I click on the browser back it works fine.  It's when I click cancel at the bottom of the wizard that I'm seeing the issue where 'Leave' requires 2 clicks to leave

indraraj:
> @indraraj - when I click on the browser back it works fine. It's when I click cancel at the bottom of the wizard that I'm seeing the issue where 'Leave' requires 2 clicks to leave

oh, I think I got it. It may be happening because clicking on "cancel" is causing both actions to trigger(hence in actual its opening two dialog box) 1)react-route URL transition and 2)Exiting wizard 

It may require some additional work for fixing this issue. @mdrillin thanks for catching it.

ui/packages/ui/src/app/pages/createConnector/CreateConnectorComponent.tsx
indraraj: yes, it can be removed.

ui/packages/ui/src/app/pages/createConnector/CreateConnectorComponent.tsx
indraraj: yeah, I missed them. I will do the clean up in the subsequent commit.

ui/packages/ui/src/app/pages/createConnector/CreateConnectorComponent.tsx
indraraj: done

ui/packages/ui/src/app/pages/createConnector/CreateConnectorComponent.tsx
indraraj: done

indraraj:
Fixed the wizard "cancel" button issue by differentiating between user navigation and browser back button navigation.

uidoyen (approved):
Tested locally. Looking nice!

Issue #278, opened 2021-01-18
Validation error message is not clear in certain scenario


uidoyen:
Scenario to reproduce:

- Select mongoDB connector then proceed  next.
- Validate  the connector basic properties.
- Under Advanced Properties panel enter 1 Seconds to `Initial delay before reconnection` and 2 Milliseconds to `Maximum delay before reconnection` then validate.

Getting the validation error message as below screenshot:

![Screenshot from 2021-01-18 15-18-04](https://user-images.githubusercontent.com/2117670/104899062-64759900-59a0-11eb-81e7-70d0d7bb71f2.png)

And API response as below:

```
[...]

PR #280, opened 2021-01-18
Handling for generic validation error

mdrillin:
Addresses Issue #278 
- added handling for GenericValidationResults in CreateConnectorComponent
- some numeric inputs require positive int - added POS-INT type and sets a different min value for those on FormInputComponent
- fix issue in CreateConnectorComponent, was checking DataOptions instead of RuntimeOptions

### DIFF
--- a/ui/packages/ui/src/app/pages/createConnector/CreateConnectorComponent.tsx
+++ b/ui/packages/ui/src/app/pages/createConnector/CreateConnectorComponent.tsx
@@ -455,19 +455,27 @@ export const CreateConnectorComponent: React.FunctionComponent<ICreateConnectorC
             getBasicPropertyDefinitions(selectedConnectorPropertyDefns),
             getAdvancedPropertyDefinitions(selectedConnectorPropertyDefns)
           );
-          for (const connectionValue of connectorPropertyDefns) {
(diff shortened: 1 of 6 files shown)


ui/packages/ui/src/app/pages/createConnector/CreateConnectorComponent.tsx
```diff
@@ -529,6 +537,7 @@ export const CreateConnectorComponent: React.FunctionComponent<ICreateConnectorC
                 }
               }
             }
+            alert(JSON.stringify(result));
```
indraraj, line 540: Need to remove the alert.

indraraj (changes requested):
UI seems to be breaking on the RuntimeOptions step for Postgres.

mdrillin:
Amended the commit to fix issues
- removed alert (oops)
- fixed problem in Utils which caused RuntimeStep error
- I also removed alert on DataOptions and RuntimeOptions step, since I was getting duplicate success messages as shown below on those steps:

DuplicateMsg

indraraj (approved):

indraraj:
Tested locally, Looks good.

PR #286, opened 2021-01-21
Adding connector actions Pause / Resume / Restart

uidoyen:
- Added connector service mock for Pause / Resume / Restart service call.
- Using `httpPost` method as of now, will replace according to backend implementation

### DIFF
--- a/ui/packages/services/src/connector/connector.service.ts
+++ b/ui/packages/services/src/connector/connector.service.ts
@@ -15,10 +15,8 @@
  * limitations under the License.
  */
 
-import {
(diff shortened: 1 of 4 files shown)


mdrillin:
After today discussion on [DBZ-2951](https://issues.redhat.com/browse/DBZ-2951) , the restart action should always be enabled - even for RUNNING and PAUSED status.  Since this is wip, I will go ahead and merge this PR.  We can fix on the next PR.

Also, we should check with UX on the connector actions - they may prefer not showing invalid actions in the menu rather than showing disabled.

mdrillin (approved):
Nice start - we can adjust for any necessary changes in the next PR

PR #287, opened 2021-01-21
Addingcolumn

indraraj:
Fixed issue https://github.com/debezium/debezium-ui-poc/issues/284 and Addresses(WIP) https://github.com/debezium/debezium-ui-poc/issues/283

- Fixed the https://github.com/debezium/debezium-ui-poc/issues/284 issue of not correctly identifying an invalid field in the case of the invalid filter expression.
- Updated the Filter page to dynamically accept an array of the possible filter field( filed name and sample value )
- Based on the passed `filedsArray` filter page is updated and the corresponding label text is adjusted.

**WIP**:-
Show the filter response tree containing column value(as per further clarification need to handle the response of `filter` API call and passed `treeData` to `FilterTreeComponent`  )
For Field type of only include/exclude to use new input component(conformation on UX)


indraraj:
I think we need not implement the first WIP task after today's call.

ui/packages/ui/src/app/pages/createConnector/connectorSteps/FilterConfigComponent.tsx
```diff
@@ -39,36 +30,9 @@ export interface IFilterConfigComponentProps {
...
-  i18nFilterExpressionResultText: string;
-  i18nCancel: string;
-  i18nClear: string;
-  i18nClearFilterConfMsg: string;
-  i18nNoMatchingTables: string;
-  i18nInvalidFilters: string;
-  i18nInvalidFilterExpText: string;
+  setIsValidFilter: (val: SetStateAction<boolean>) => void;
+  selectedConnectorType: string;
+  filtersArray: string[];
```
mdrillin, line 35: filtersArray isn't used - you can remove it.  And then no need to pass in from CreateConnectorComponent.

ui/packages/ui/src/app/components/FilterInputFieldComponent.tsx
```diff
@@ -0,0 +1,207 @@
...
+    getFieldExpression(
+      props.filterValues,
+      props.fieldExcludeList,
+      props.fieldIncludeList
+    )
+  );
+
+  const [fieldSelected, setFieldSelected] = React.useState<string>(
+    props.filterValues.has(props.fieldExcludeList)
+      ? "filedExclude"
```
mdrillin, line 69: minor misspelling.  should be "fieldExclude" / "fieldInclude" ?

mdrillin (approved):
Tested locally - looks really good!  Looks like filtersArray property in FilterConfigComponent is not needed though.  If you want to make couple changes first - go ahead and then merge

ui/packages/ui/src/app/components/FilterInputFieldComponent.tsx
indraraj: done

ui/packages/ui/src/app/pages/createConnector/connectorSteps/FilterConfigComponent.tsx
indraraj: done

Issue #290, opened 2021-01-22
Filter Screen - Result error page persists, doesn't clear on success


mdrillin:
On the filter screen in the wizard, an invalid filter entry will sometimes result in "Internal Server Error" in the results area. (See screenshot below).  Correcting the filter value and re-apply does not clear the error / display valid results.  It would be nice to display something more user-friendly in that area (instead of the generic server error).  And subsequent successful re-apply of the filter should dismiss the error and show valid results...

To reproduce
1) enter "inventory" in the field exclusion filter and apply (get error)
2) enter "inventory.collection.field" in the field exclusion filter and apply (no service error but page is unchanged)

![FilterServerError](https://user-images.githubusercontent.com/1820403/105516070-1069fd80-5c9b-11eb-877d-b65e5caa6a55.png)

mdrillin:
Pending PR #291

indraraj:
Pr merged.

PR #291, opened 2021-01-22
addresses filter result errors

mdrillin:
Address issue #290 
- FilterConfigComponent now clears any errors when filters cleared or applied
- changes in ApplicationErrorPage to i18n some text, better display of error details

### DIFF
--- a/ui/packages/ui/src/app/components/ApplicationErrorPage.tsx
+++ b/ui/packages/ui/src/app/components/ApplicationErrorPage.tsx
@@ -8,11 +8,12 @@ import {
   PageSection,
   PageSectionVariants,
   Text,
-  Title,
+  Title
 } from "@patternfly/react-core";
 import { ExclamationTriangleIcon } from "@patternfly/react-icons";
-import { useState } from "react";
 import * as React from "react";
+import { useState } from "react";
+import { useTranslation } from 'react-i18next';
 import "./ApplicationErrorPage.css";
 
 export interface IApplicationErrorPageProps {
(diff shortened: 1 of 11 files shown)


mdrillin:
@indraraj - after working on this, I'm thinking we could potentially eliminate usage of WithLoader in FilterTreeComponent?  Then we wouldn't display that whole ApplicationErrorPage when we get an api error here.  wdyt?

ui/packages/ui/src/app/components/ApplicationErrorPage.tsx
```diff
@@ -8,60 +8,72 @@ import {
...
 import { ExclamationTriangleIcon } from "@patternfly/react-icons";
-import { useState } from "react";
 import * as React from "react";
+import { useState } from "react";
+import { useTranslation } from 'react-i18next';
 import "./ApplicationErrorPage.css";
 
 export interface IApplicationErrorPageProps {
   icon?: React.ComponentType<any>;
-  title?: string;
```
indraraj, line 20: The idea to have these optional props was to make this component configurable and use in diff error causing scenarios, whereas the default case will be used in `WithErrorBoundary` component

ui/packages/ui/src/app/components/ApplicationErrorPage.tsx
```diff
@@ -8,60 +8,72 @@ import {
...
+
   return (
     <React.Fragment>
       <PageSection className="ps_error" variant={PageSectionVariants.light}>
         <div className="application-error-page">
           <EmptyState variant={EmptyStateVariant.large}>
             <EmptyStateIcon icon={props.icon || ExclamationTriangleIcon} />
             <Title headingLevel="h5" size="lg">
-              {props.title || "Something is wrong"}
+              {t('appErrorTitle')}
```
indraraj, line 52: This will show server error on all the scenarios even in case of UI breaking.

indraraj:
> @indraraj - after working on this, I'm thinking we could potentially eliminate usage of WithLoader in FilterTreeComponent? Then we wouldn't display that whole ApplicationErrorPage when we get an api error here. wdyt?

Sorry, I'm afraid I am not able to follow you here on why to remove `WithLoader`, and if we want to display something else on API error we can do it  by passing custom `errorChildren` e.g ` errorChildren={<ApiError error={props.errorMsg} />}`

mdrillin:
Yeah true on the WithLoader comment - better alternate to use custom errorChildren.  I can address the couple comments and update

mdrillin:
Updated the PR to address comments

indraraj (approved):
Tested locally, looks nice.

PR #297, opened 2021-01-28
Filter alerts

indraraj:
- Added the column/field filter being applied info in the filter page alerts.
- Moved the filter page alert from `FilterConfigComponent` to `FilterTreeComponent` to remove the multiple spinners.
screenshot 
image


mdrillin (approved):
lgtm

PR #298, opened 2021-01-28
Adding display of connector type within wizard pages

mdrillin:
- Adds display of the connector type on the basic properties page, underneath the name input (see images below)
- Adds header showing the connector name/type on the subsequent wizard pages (see images below)
- changed activeStep.name comparisons in CreateConnectorComponent to activeStep.id comparison.  Think this would be more reliable, since some of the names are actually elements - not strings.

**_Connection Props step:_**
PropertiesStep

**_Subsequent wizard steps:_**
OtherPropsStep

### DIFF
--- a/ui/packages/ui/src/app/pages/createConnector/CreateConnectorComponent.tsx
+++ b/ui/packages/ui/src/app/pages/createConnector/CreateConnectorComponent.tsx
@@ -50,6 +50,7 @@ import {
   ReviewStepComponent,
   RuntimeOptionsComponent,
 } from "./connectorSteps";
+import { ConnectorNameTypeHeader } from './connectorSteps/ConnectorNameTypeHeader';
 import "./CreateConnectorComponent.css";
 
 /**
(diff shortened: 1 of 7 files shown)


indraraj (approved):
Tested locally looks good.

indraraj:
Hey @mdrillin have a look at the screenshot below I was thinking if we should display it like this. 
image

We can avoid the `Connector name` & `Type` labels as the created connector on the `Connector` page looks something like this (connector type is being identified used DB image) and secondly, on the wizard top it looks more clear 
image

Anyway, designs are always subjective. I have approved the current pr you can merge it if you like.

mdrillin:
@indraraj  - yeah, I did consider putting the info in the 'header' area where you are suggesting.  It does look a little more concise that way, but I'd rather have it more to the left (seems out of place to the right).  I think I'll go ahead and merge what we have here and ask for UX opinion on the placement.  We can change it later if needed.

PR #302, opened 2021-02-01
Making connector list items expandable

uidoyen:
- Implemented connector items expandable.
- Added total task count based on task status.
- Added individual task restart and popover.

Connector overview message is static for now.

### DIFF
--- a/ui/packages/ui/src/app/pages/connectors/ConnectorTask.tsx
+++ b/ui/packages/ui/src/app/pages/connectors/ConnectorTask.tsx
@@ -11,8 +11,8 @@ export interface IConnectorTaskProps {
   i18nRestart: string;
   i18nTaskStatusDetail: string;
   i18nTaskErrorTitle: string;
-  setConnectorToRestart: (connName: string) => void;
-  showRestartConfirmationDialog: () => void;
+  connectorTaskToRestart: (connName: string, taskId: string) => void;
+  showConnectorTaskToRestartDialog: () => void;
 }
 
 /**
@@ -44,9 +44,9 @@ if (props.errors) {
     errors.push(<div key ={index} className="connector-task-error">{error}</div>)
   });
 }
-const doRestart = (connName) => {
-  props.setConnectorToRestart(connName);
-  props.showRestartConfirmationDialog();
+const doConnectorTaskRestart = (connName: string, taskId: string) => {
+  props.connectorTaskToRestart(connName, taskId);
+  props.showConnectorTaskToRestartDialog();
 }
 
 return (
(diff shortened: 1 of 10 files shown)


ui/packages/ui/src/app/pages/connectors/ConnectorTask.tsx
```diff
@@ -1,15 +1,18 @@
...
 export interface IConnectorTaskProps {
   status: string;
+  connName: string;
   taskId: string;
   errors?: any;
   i18nTask: string;
+  i18nRestart: string;
   i18nTaskStatusDetail: string;
+  i18nTaskErrorTitle: string;
+  setConnectorToRestart: (connName: string) => void;
```
mdrillin, line 14: The specific connector task should be restarted.  So this should be something like "setConnectorTaskToRestart: (connName: string, taskId: number)..."

ui/packages/ui/src/app/pages/connectors/ConnectorsTableComponent.tsx
```diff
@@ -183,18 +192,24 @@ export const ConnectorsTableComponent: React.FunctionComponent<IConnectorsTableC
...
           key={id}
           status={taskState.taskStatus}
+          connName={conn.name}
           taskId={id}
           errors={taskState.errors}
           i18nTask={t('task')}
+          i18nRestart={t('restart')}
           i18nTaskStatusDetail={t('taskStatusDetail')}
+          i18nTaskErrorTitle={t('taskErrorTitle')}
+          setConnectorToRestart={connName}
```
mdrillin, line 211: should be 'setConnectorTaskToRestart', which would initiate restart of the designated task number for the connector

ui/packages/ui/src/app/pages/connectors/ConnectorTask.tsx
```diff
@@ -1,15 +1,18 @@
...
+  showRestartConfirmationDialog: () => void;
```
mdrillin, line 15: There will need to be a separate confirmation which indicates restart of connector name, task id

mdrillin (changes requested):
I left a couple comments.  Looks like a really good start on the expandable connector rows!
- I believe the task restart should restart the specific connector/task and not the connector.  This means the task id will need to be supplied along with the connector name.  And initiate a different call to the backend service for the task restart.
- Also noticed that if you expand the connector row, it will collapse within a short time when the auto-refresh happens

mdrillin (approved):
looks good - we can address the auto-refresh issue in a separate PR

PR #303, opened 2021-02-01
Update backend connector service calls

mdrillin:
- updates the connector state lifecycle service methods and adds the task method
- update the restart action enablement - should always be enabled
- add function in ConnectorsTableComponent which changes a connector state immediately upon success of the lifecycle change instead of waiting for periodic refresh


uidoyen (approved):
Tested locally. Nice progress!

PR #304, opened 2021-02-02
moving the column/field filter under show/hide button

indraraj:
Implemented the fix for https://github.com/debezium/debezium-ui-poc/issues/301


mdrillin (approved):
LGTM

PR #306, opened 2021-02-02
persists connectorsTable sortCategory

mdrillin:
Address issue where the connectorsTable sortCategory is lost if user moves to a different page, and then comes back.  Uses localStorage to initialize and maintain the sortCategory selection.  Similar approach could be used for the sortDirection and also the row expansion states.


mdrillin:
Closing, out of sync - may resubmit later

PR #309, opened 2021-02-03
connectors table updates

mdrillin:
- fix connector service restart functions - POST instead of PUT
- consolidations in ConnectorsTableComponent
    - using single ConfirmationDialog and setting properties for the current action
    - separately managed states for the various actions are consolidated
    - separate functions for each action were consolidated into doPerformCurrentAction with switch-case
    - added connector name / task info to confirmation dialog messages
- change sort icons to agree with patternfly ascending / descending guideline https://www.patternfly.org/v4/guidelines/icons


uidoyen (approved):
Nice update!

PR #310, opened 2021-02-08
DBZ-3054 Improve connector type display

mdrillin:
Simplified display of connector type in the wizard, per UX request.
- type icon is no longer shown (added 'showIcon' property to components to allow hide or show)
- simplified the name / type displayed at top of wizard pages to remove the bold labels, added divider

**Properties step screenshot:**
props-cType

**Steps 3+ screenshot:**
subsequent-nameType


indraraj (approved):
Tested locally, looks good.

PR #311, opened 2021-02-09
fix auto expanding issue

indraraj:
Fixed issue [DBZ-3049](https://issues.redhat.com/browse/DBZ-3049)

- Converted `expandableRows` state to an array type, with value storing all the expanded row index.
- Fixed the value provided to `isOpen` props in the table row (`isOpen: expandedRows.includes(index*2),`)


mdrillin (approved):
Tested several scenarios locally - looks good!

PR #314, opened 2021-02-10
added white-space to pre tag in server error msg detail

indraraj:
Added the white-space to the `pre` tag so that the error detail msg does not print in long-line. 
screenshot
image


mdrillin (approved):
lgtm

PR #315, opened 2021-02-10
DBZ-3091 Adding UI props for MySQL connector type

mdrillin:
- Added new property names for MySQL connector type
- re-organized property names into alpha order


uidoyen (approved):
Looks good!

PR #316, opened 2021-02-10
DBZ-3027 - Remove POC references from UI

mdrillin:
Initial changes for removing POC references.  The repo name will need to be changed to 'debezium-ui'.  Also the dockerhub image 'debezium-ui-poc'

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -3,7 +3,7 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>io.debezium.ui</groupId>
+    <groupId>io.debezium</groupId>
     <artifactId>debezium-ui-parent</artifactId>
     <version>1.0.0-SNAPSHOT</version>
   </parent>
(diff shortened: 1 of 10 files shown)


backend/pom.xml
```diff
@@ -3,12 +3,12 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
   <modelVersion>4.0.0</modelVersion>
   <parent>
-    <groupId>io.debezium.uipoc</groupId>
-    <artifactId>debezium-uipoc-parent</artifactId>
+    <groupId>io.debezium.ui</groupId>
```
rk3rn3r, line 6: @gunnarmorling should we align that groupId to the package namespace of the Java application?

rk3rn3r (approved):
LGTM

mdrillin:
@rk3rn3r does your question to @gunnarmorling need an answer, or should we merge this?  I was going to merge then rename the repo - but looks like the rename has already been done.

backend/pom.xml
gunnarmorling: Or maybe even io.debezium, like everything else? I don't think it needs a separate group id actually.

rk3rn3r: Makes sense. Let's do that @mdrillin

rk3rn3r:
@mdrillin can you make the above change to the maven group id, then we are good to go

gunnarmorling (approved):

gunnarmorling:
Thx, @mdrillin!

PR #317, opened 2021-02-11
null check on ui for weeding out backend null prop value

indraraj:
Null check on UI for defective properties value(null)


mdrillin (approved):
looks good

PR #318, opened 2021-02-11
DBZ-3091 Adds MySQL property handling in UI Utils

mdrillin:
Adds layout formatting for MySQL properties to Utils function.  Also addition of some missing property names, corrections and improvements.


uidoyen (approved):
Tested locally. Looks good!

PR #320, opened 2021-02-15
DBZ-3105 Layout improvement

uidoyen:
- Improve layout for Data options component, switch buttons overlapping on mobile
- Reduced main wizard sub nav item gape
- Fixed layout for Column Truncate in Mask Component


mdrillin (approved):
LGTM

PR #324, opened 2021-02-18
DBZ-3052 Adding build commit sha with link to changes.

brbrown25:


brbrown25:
@gunnarmorling @mdrillin First pass at adding in the commit sha

image

gunnarmorling:
Thanks a lot, @brbrown25!

@rk3rn3r @mdrillin, could you take a look at this one?

ui/packages/ui/webpack.common.js
```diff
@@ -6,6 +6,7 @@ const { dependencies } = require("./package.json");
 
 // Try the environment variable, otherwise use root
 const ASSET_PATH = process.env.ASSET_PATH || '/';
+const COMMIT_HASH = require('child_process').execSync('git rev-parse --short HEAD').toString().trim();
```
gunnarmorling, line 9: Probably a real noob question: but this runs at _build-time_, right? I.e. there's no assumption that this is run from within the source repo, correct?

ui/packages/ui/src/app/Layout/AppLayout.tsx
```diff
@@ -21,7 +21,13 @@ const AppLayout: React.FC = ({ children }) => {
 	);
 	const handleClusterIdChange = (id: number) => {
 		setClusterId(id);
-	}
+	};
+	const commitLink = `https://github.com/debezium/debezium-ui/commit/${process.env.COMMIT_HASH}`;
+	const Footer = () => (
+  	<footer class="footer">
```
uidoyen, line 27: Should be `className` instead of `class`

uidoyen (changes requested):

uidoyen:
Looks like footer is overlapping main Wizard buttons.
Screenshot from 2021-02-18 13-46-44

ui/packages/ui/src/app/Layout/AppLayout.tsx
brbrown25: :doh: I forgot about that, my react/ts is a little rusty.

ui/packages/ui/webpack.common.js
brbrown25: Not a noob question at all! I believe that is how this works, it should get evaluated when `start` or `build` are invoked. according to the docs 

> The DefinePlugin allows you to create global constants which can be configured at compile time.
https://webpack.js.org/plugins/define-plugin/#root

brbrown25:
- [ ] fix overlapping footer
- [ ] class -> className

brbrown25:
I'll knock these out today. thanks for the speed review everyone!

brbrown25:
@uidoyen it's a little closer, no longer overlapping but still not quite where I feel like it should be. I'm not that familiar with patternfly so not sure if there is a better way to get the footer placed, open to suggestions.

mdrillin:
@brbrown25 thanks for submitting this!  Sorry I didnt pick up on it before, but I think the patternfly design recommendation is to use an about dialog instead of putting the build info in the footer.  I added a [comment to the JIRA](https://issues.redhat.com/browse/DBZ-3052?focusedCommentId=15810159&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15810159) .  @gunnarmorling you ok with that approach?

brbrown25:
> @brbrown25 thanks for submitting this! Sorry I didnt pick up on it before, but I think the patternfly design recommendation is to use an about dialog instead of putting the build info in the footer. I added a [comment to the JIRA](https://issues.redhat.com/browse/DBZ-3052?focusedCommentId=15810159&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15810159) . @gunnarmorling you ok with that approach?

no problemo, thanks for the opportunity! Once we agree on next steps I'll give this an update.

gunnarmorling:
> @gunnarmorling <https://github.com/gunnarmorling> you ok with that
approach?

Yes, absolutely. Let's stick to the Patternfly recommendations wherever
they exist and they make sense for us.

>

brbrown25:
I'll knock this out tomorrow morning!

brbrown25:
@mdrillin I've pulled in the latest master and switched to a modal, i didn't play with the sizing our anything. another option could be to have a tooltip that is activated by a click. although I think this modal approach is nice since there is some additional information that could be added over time

ui/packages/ui/src/app/Layout/appHeader.tsx
```diff
@@ -18,10 +23,33 @@ const AppHeader: React.FC<IAppHeader> = (props) => {
...
 		history.push("/")
 	}
-
+	const [isModalOpen, setIsModalOpen] = React.useState(false);
+	const handleModalToggle = () => setIsModalOpen(!isModalOpen);
+	const commitLink = `https://github.com/debezium/debezium-ui/commit/${process.env.COMMIT_HASH}`;
+	const Footer = () => (
+		<p>Build <a href={commitLink} target="_blank">{process.env.COMMIT_HASH}</a></p>
+	);
+	const BuildModal = () => (
```
brbrown25, line 32: I was tempted to put this in it's own file but wasn't sure what the preferences are for that.

ui/packages/ui/src/app/Layout/appHeader.tsx
```diff
@@ -18,10 +23,33 @@ const AppHeader: React.FC<IAppHeader> = (props) => {
 	const homeClick = () => {
 		history.push("/")
 	}
-
+	const [isModalOpen, setIsModalOpen] = React.useState(false);
+	const handleModalToggle = () => setIsModalOpen(!isModalOpen);
+	const commitLink = `https://github.com/debezium/debezium-ui/commit/${process.env.COMMIT_HASH}`;
+	const Footer = () => (
```
mdrillin, line 29: This Footer can be removed - no longer needed.

ui/packages/ui/src/app/Layout/appHeader.tsx
mdrillin: I think it's fine for now.  We can move it later if we start adding more info.

mdrillin:
I think this looks good.  Please remove Footer in appHeader, then we are good to go.  Thanks!

ui/packages/ui/src/app/Layout/appHeader.tsx
brbrown25: opps, I totally thought I removed that!

brbrown25:
@mdrillin thanks so much! I've collapsed things down and removed `Footer`

mdrillin (approved):
Changes look good - thanks @brbrown25 !

PR #332, opened 2021-03-17
DBZ-3265 using plugin to obtain git revision

mdrillin:
utilizing plugin to obtain the git commit hash for the build

### DIFF
--- a/ui/packages/ui/webpack.common.js
+++ b/ui/packages/ui/webpack.common.js
@@ -3,10 +3,12 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
 const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
 const webpack = require('webpack');
 const { dependencies } = require("./package.json");
+const GitRevisionPlugin = require('git-revision-webpack-plugin');
+const gitRevisionPlugin = new GitRevisionPlugin({commithashCommand: 'rev-parse --short HEAD'});
 
 // Try the environment variable, otherwise use root
 const ASSET_PATH = process.env.ASSET_PATH || '/';
-const COMMIT_HASH = process.env.COMMIT_HASH || require('child_process').execSync('git rev-parse --short HEAD').toString().trim();
+const COMMIT_HASH = process.env.COMMIT_HASH || gitRevisionPlugin.commithash();
 
 module.exports = {
   entry: {
@@ -20,8 +22,6 @@ module.exports = {
     // This makes it possible for us to safely use env vars on our code
     new webpack.DefinePlugin({
       'process.env.ASSET_PATH': JSON.stringify(ASSET_PATH),
-    }),
-    new webpack.DefinePlugin({
       'process.env.COMMIT_HASH': JSON.stringify(COMMIT_HASH),
     }),
     new webpack.container.ModuleFederationPlugin({
--- a/ui/package.json
+++ b/ui/package.json
@@ -29,6 +29,7 @@
     }
   },
   "devDependencies": {
+    "git-revision-webpack-plugin": "^3.0.6",
     "husky": "^4.3.7",
     "lerna": "3.21.0"
   },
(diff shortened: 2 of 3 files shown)


indraraj (approved):
Tested local and container builds, looks fine to me.

ui/packages/ui/webpack.common.js
```diff
@@ -3,10 +3,12 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
 const TsconfigPathsPlugin = require('tsconfig-paths-webpack-plugin');
 const webpack = require('webpack');
 const { dependencies } = require("./package.json");
+const GitRevisionPlugin = require('git-revision-webpack-plugin');
+const gitRevisionPlugin = new GitRevisionPlugin({commithashCommand: 'rev-parse --short HEAD'});
```
rk3rn3r, line 7:
```suggestion
const gitRevisionPlugin = new GitRevisionPlugin({versionCommand: 'describe --always --tags'});
```

ui/packages/ui/webpack.common.js
```diff
@@ -3,10 +3,12 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
...
 
 // Try the environment variable, otherwise use root
 const ASSET_PATH = process.env.ASSET_PATH || '/';
-const COMMIT_HASH = process.env.COMMIT_HASH || require('child_process').execSync('git rev-parse --short HEAD').toString().trim();
+const COMMIT_HASH = process.env.COMMIT_HASH || gitRevisionPlugin.commithash();
```
rk3rn3r, line 11:
```suggestion
const COMMIT_HASH = process.env.COMMIT_HASH || gitRevisionPlugin.version();
```

rk3rn3r: We can capture the latest tag (that build is based on, plus the short hash) like this:
```
$ git describe --always --tags
v1.5.0.Beta2-2-g3cf30f2
```

ui/packages/ui/webpack.common.js
```diff
@@ -20,8 +22,6 @@ module.exports = {
     // This makes it possible for us to safely use env vars on our code
     new webpack.DefinePlugin({
       'process.env.ASSET_PATH': JSON.stringify(ASSET_PATH),
-    }),
-    new webpack.DefinePlugin({
       'process.env.COMMIT_HASH': JSON.stringify(COMMIT_HASH),
```
rk3rn3r, line 25:
```suggestion
      'process.env.COMMIT_VERSION': JSON.stringify(COMMIT_VERSION),
```

ui/packages/ui/webpack.common.js
rk3rn3r: What happens here, when git is not available? Did you test that (removing `git` from the `microdnf install git` call in the builder part of the Dockerfile? Do we have a proper error message?

uidoyen: It looks like we need `git` installed (see https://github.com/pirelenito/git-revision-webpack-plugin/blob/v3/lib/helpers/run-git-command.js#L9)

rk3rn3r: I know. But the question is what happens when it's not available. I expect the build to fail, and not, like before, continue without assets.

uidoyen: I tried after removing `git` from  `Dockerfile` and build was success this time.

rk3rn3r: And what is the COMMIT_HASH then or is it just empty? This should not happen. When the hash (and git) is unavailable the build should fail. WDYT?

uidoyen: Yes, the build should fail when `git` is unavailable but build is success and I am getting`COMMIT_HASH` is `3762ffd`.

rk3rn3r: @uidoyen So what is your local `git rev-parse --short HEAD`? Is it the correct value? And why does it "work"?

indraraj:
@mdrillin I think this can be closed since @uidoyen  is working to use the GitHub API to fetch the commit id which don't need git to be locally installed.

PR #334, opened 2021-03-23
DBZ-3316 Remove properties from connector-types response

mdrillin:
backend:
connector-types - removes properties from the response
connector-types/{id} - now only returns the array of properties

ui:
remove properties attribute from the ConnectorType model
adjust the service and usages to account for the connector-types/{id} only returning properties array

### DIFF
--- a/backend/src/main/java/io/debezium/configserver/model/ConnectorDefinition.java
+++ b/backend/src/main/java/io/debezium/configserver/model/ConnectorDefinition.java
@@ -1,0 +1,27 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.configserver.model;
+
+public class ConnectorDefinition {
+
+    public String id;
+    public String className;
+    public String displayName;
+    public String version;
+    public boolean enabled;
+
+    public ConnectorDefinition() {
+    }
+
+    public ConnectorDefinition(String id, String className, String displayName, String version, boolean enabled) {
+        this.id = id;
+        this.className = className;
+        this.displayName = displayName;
+        this.version = version;
+        this.enabled = enabled;
+    }
+
+}
--- a/backend/src/main/java/io/debezium/configserver/model/ConnectorType.java
+++ b/backend/src/main/java/io/debezium/configserver/model/ConnectorType.java
@@ -7,24 +7,15 @@ package io.debezium.configserver.model;
 
 import java.util.List;
 
-public class ConnectorType {
+public class ConnectorType extends ConnectorDefinition {
 
-    public String id;
-    public String className;
-    public String displayName;
-    public String version;
-    public boolean enabled;
     public List<ConnectorProperty> properties;
 
     public ConnectorType() {
     }
 
     public ConnectorType(String id, String className, String displayName, String version, boolean enabled, List<ConnectorProperty> properties) {
-        this.id = id;
-        this.className = className;
-        this.displayName = displayName;
-        this.version = version;
-        this.enabled = enabled;
+        super(id, className, displayName, version, enabled);
         this.properties = properties;
     }
 }
--- a/backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
+++ b/backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
@@ -42,6 +42,7 @@ import javax.ws.rs.core.Response.Status;
 import io.debezium.DebeziumException;
 import io.debezium.configserver.model.ConnectionValidationResult;
 import io.debezium.configserver.model.ConnectorType;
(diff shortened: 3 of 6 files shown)


backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
```diff
@@ -102,7 +103,7 @@ public Response getClusters() throws InvalidClusterException {
             responseCode = "200",
             content = @Content(
                     mediaType = MediaType.APPLICATION_JSON,
-                    schema = @Schema(implementation = ConnectorType.class)
+                    schema = @Schema(implementation = ConnectorProperty.class, type = SchemaType.ARRAY)
```
rk3rn3r, line 106: Like in your task comment, I expected that we keep `ConnectorType.class` here. (See my email and let's discuss it in today's standup call).

rk3rn3r:
As discussed we will implement your suggested solution from the task description:

api/connector-types  - remains as is, (without the properties).

/api/connector-types/{id} - return same info for a single type (including properties)

mdrillin:
Pushed update after today's discussion.  /api/connector-types remains same (except without properties).  /api/connector-types/{id} returns same info for a single connector, and includes properties

indraraj (approved):
Looks good from the UI side.

backend/src/main/java/io/debezium/configserver/model/ConnectorType.java
```diff
@@ -5,26 +5,23 @@
...
     public String displayName;
     public String version;
     public boolean enabled;
-    public List<ConnectorProperty> properties;
 
     public ConnectorType() {
     }
 
-    public ConnectorType(String id, String className, String displayName, String version, boolean enabled, List<ConnectorProperty> properties) {
+    public ConnectorType(String id, String className, String displayName, String version, boolean enabled) {
```
rk3rn3r, line 19: In `io.debezium.configserver.service.ConnectorIntegrator` the method that returns instances of this class is called `getDescriptor()`.  WDYT @mdrillin about renaming this class to `ConnectorDescriptor`?

backend/src/main/java/io/debezium/configserver/model/ConnectorTypeFull.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.configserver.model;
+
+import java.util.List;
+
+public class ConnectorTypeFull extends ConnectorType {
```
rk3rn3r, line 10: `Full` feels a bit unspecific, especially as a suffix which is often used to make things more specific (but "full" is not specific) and related to renaming the original `ConnectorType` class to `ConnectorDescriptor` we could maybe keep this name as `ConnectorType`? Or maybe vice versa. I am also fine with a good new name for both.

I could also live with `ConnectorTypeWithProperties`, but I think this class is THE thing what a connector type should be. So the other one must be something different.

rk3rn3r:
Thx @mdrillin. The change looks good. My only & minor concern is the naming. You don't have to do the renaming, but I'm interested in your opinion.

backend/src/main/java/io/debezium/configserver/model/ConnectorTypeFull.java
mdrillin: @rk3rn3r yeah sometimes it's a struggle to come up with appropriate names...
So maybe we could use ConnectorType for 'no properties' and ConnectorDefinition for 'with properties'?

Then the ConnectorIntegrator interfaces would be:

ConnectorType getConnectorType();
ConnectorDefinition getConnectorDefinition();

wdyt?

mdrillin:
I went ahead and pushed an update for the naming changes

backend/src/main/java/io/debezium/configserver/model/ConnectorDefinition.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.configserver.model;
+
+import java.util.List;
+
+public class ConnectorDefinition extends ConnectorType {
```
rk3rn3r, line 10: I'm sorry @mdrillin. This is still the wrong direction.
A `ConnectorType` should be the FULL thing including properties. As it was before.

A `SomeOtherThing`, for example `ConnectorDescriptor` is without properties.

mdrillin: In my mind 'ConnectorType' is a smaller set of info than a 'ConnectorDescriptor'.  But no problem, if you guys want it the opposite way - I can change it.

PR #343, opened 2021-05-04
Handling the filter step data and passing it to configuration in host…

indraraj:
… app


uidoyen (approved):
Tested locally. Looks good.

PR #345, opened 2021-05-05
Adapting to generated schema properties

mdrillin:
Adding the generated properties mocks and adapting DebeziumConfigurator
- changed getFormattedProperties in Utils - arg is now just the connector typeId.  adapted the usages.
- updated getPropertiesData in DebeziumConfigurator to utilize the generated properties mock


uidoyen (approved):
Tested locally. Looks good.

PR #346, opened 2021-05-07
Fixing FormDurationComponent & FormSelectComponent issue

uidoyen:
Fixing FormDurationComponent & FormSelectComponent issues

### DIFF
--- a/ui/packages/ui/src/app/components/formHelpers/FormSelectComponent.tsx
+++ b/ui/packages/ui/src/app/components/formHelpers/FormSelectComponent.tsx
@@ -64,7 +64,7 @@ export const FormSelectComponent = (props: IFormSelectComponentProps) => {
   };
 
   const selectOptions = options.map((value: any) => {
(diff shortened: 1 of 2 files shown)


ui/packages/ui/src/app/components/formHelpers/FormSelectComponent.tsx
```diff
@@ -64,7 +64,7 @@ export const FormSelectComponent = (props: IFormSelectComponentProps) => {
   };
 
   const selectOptions = options.map((value: any) => {
-    return {'value': value.charAt(0).toUpperCase() + value.slice(1)}
+    return {'value': value.charAt(0) + value.slice(1)}
```
mdrillin, line 67: I don't think you need to split the first character out anymore - just return the value?

mdrillin (approved):
tested and is now working fine.  I had one comment on FormSelectComponent - if you want to look at that - then go ahead and merge.  Thanks!

PR #347, opened 2021-05-10
Fix required property messages; other minor fixes

mdrillin:
- for the required properties, changed the message to include property name - same as standalone
- replaced hardcoded invalid message
- renamed some component properties for connectorType
- DebeziumConfigurator passes connectorType into steps (still hardcoded to the postgres schema)
- removed some unused component properties
- name configurator steps to match standalone app


uidoyen (approved):
Tested locally. Looks good.

indraraj (approved):
Tested locally, looks good.

PR #354, opened 2021-05-19
DBZ-3102: Disable the toggle if no value is entered in input field

indraraj:
Issue:- https://issues.redhat.com/browse/DBZ-3102


mdrillin:
I put a comment on the JIRA - I'm thinking maybe just disable the toggle group if there is no value?  wdyt?

indraraj:
@mdrillin I have updated the PR. 
Since patternFly does not have a `disable` option for `ToggleGroup` i had to disable the `ToggleGroupItem` individually. Also added this to their issue. https://github.com/patternfly/patternfly-react/issues/5809

mdrillin (approved):
looks good - thanks!

PR #360, opened 2021-06-01
DBZ-3580 - handle nulls in connectors response

mdrillin:
Filtering out nulls from connectors list in the UI - in case the backend response has nulls mixed in.


indraraj (approved):
Tested locally with mock api seems to be fixing the issue.

PR #365, opened 2021-06-21
Remove name mcs

indraraj:
- Break-fix and removing the unused props.


uidoyen (approved):
Working fine. Thanks!

PR #366, opened 2021-06-24
Fix for required properties with default values

indraraj:
Add the required properties with default values in the final configurator JSON, even if values are not modified.


mdrillin (approved):
looks good!

PR #377, opened 2021-07-09
Adding more component tests

mdrillin:
Adding tests for more of the lower level components.  They are mostly just basic tests to render the components - we can improve the coverage over time.


indraraj (approved):
Looks good, thanks for working on it.

PR #378, opened 2021-07-15
DBZ-3265 catch build error to exported checkout which has no .git dir

uidoyen:
- Catching build error with proper msg when .git dir isn't available
- Removed `git-revision-webpack-plugin` which is no longer in use

Note: Build is successful even if  `.git` dir isn't available but in this case build info can't be seen in UI. 
Test case: Try removing `.git` dir and build it.

Screenshot from 2021-07-15 11-41-42


ui/packages/ui/webpack.common.js
```diff
@@ -39,7 +42,6 @@ module.exports = (argv) => {
         {
           test: /\.css$/,
           use: [
-            // MiniCssExtractPlugin.loader,
```
indraraj, line 42: I had intentionally left this commented to give the bug reproduction scenario. here https://github.com/patternfly/patternfly-react/issues/5726

indraraj (approved):
apart from above comment rest lgtm

ui/packages/ui/webpack.common.js
uidoyen: Added it back.

PR #381, opened 2021-07-29
DBZ-3794: Allowing the wizard main section to be scrollable to handle…

indraraj:
… the any runtime element attached


uidoyen (approved):
LGTM!

PR #382, opened 2021-07-29
DBZ-3783 Disable next upon form change

uidoyen:


indraraj:
@uidoyen while testing this I found one more issue described below, can you also handle it while you are on this. Thanks 

In case of a duplicate name issue, the error msg for the input box does not clear up on changing the input. Whereas pager level msg get disappear. Input box error status should also be removed on change. screenshot below.

image

image

uidoyen:
Fixed as per the comment.

indraraj (approved):
Looks good, thanks

PR #385, opened 2021-08-06
Dbz 3830

indraraj:
- Using Patternfly `Code` component to show the final JSON instead of textarea.
- Added option to download the final JSON or copy to clipboard.
- Removed the titles from the steps page.


uidoyen (approved):
LGTM

PR #388, opened 2021-08-11
adding option to view the password on review page

indraraj:
Screenshot
image


indraraj:
> The mongodb password doesn't get masked - can you include that also? (the key is different for mongo)

Done

mdrillin (approved):
lgtm

PR #389, opened 2021-08-11
DBZ-3698 SMT

indraraj:


indraraj:
I will be raising a separate PR for cypress test cases.

uidoyen (approved):
Tested locally. Looks good!

indraraj:
> @mdrillin @uidoyen FYI There is one known issue with this PR that I will be handling in subsequent PR, described below
> 
> `enableReinitialize` of Formik forms is not working correctly so after the form is rendered first time and if you change to load diff form its initial value are not show in UI.
> 
> Apart from above issue reset should work correctly.

Fixed in this PR https://github.com/debezium/debezium-ui/pull/390

PR #390, opened 2021-08-31
Fixed open issue and added more dbz SMT

indraraj:


ui/packages/ui/src/app/components/formHelpers/FormInputComponent.tsx
```diff
@@ -66,7 +66,7 @@ export const FormInputComponent: React.FunctionComponent<IFormInputComponentProp
           field.onChange(field.name)(e);
           props.clearValidationError();
         }}
-        defaultValue={field.value}
+        value={field.value}
```
indraraj, line 69: @uidoyen i know you have worked on some fixes around this, let me know if its breaking any thing.

mdrillin: I'm seeing a console error which is probably related to this.  (Happens on connector properties when validating)

Warning: A component is changing a controlled input to be uncontrolled. This is likely caused by the value changing from a defined to undefined, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://reactjs.org/link/controlled-components
    at input
    at TextInputBase (webpack-internal:///../../node_modules/@patternfly/react-core/dist/esm/components/TextInput/TextInput.js:37:9)
    at TextInput

indraraj: Done, thanks for catching that.

mdrillin (approved):
Looks good, however I do see an issue with the 'ByLogicalTableRouter' transform.  You can open a separate JIRA for it if you prefer, or commit another fix to this before merging.  If you add 2 'ByLogicalTableRouter' transforms, then toggle the 'Enforce key uniqueness' property on the second transform - it changes that property on the first transform...

indraraj:
> Looks good, however I do see an issue with the 'ByLogicalTableRouter' transform. You can open a separate JIRA for it if you prefer, or commit another fix to this before merging. If you add 2 'ByLogicalTableRouter' transforms, then toggle the 'Enforce key uniqueness' property on the second transform - it changes that property on the first transform...

thanks @mdrillin I will open a separate Jira for this one

PR #393, opened 2021-09-08
DBZ-3967 Fix issue with removal of default properties

mdrillin:
Fixes issue with removal properties that have defaults.  The check in Utils.minimizePropertyValues was incorrect.

### DIFF
--- a/ui/packages/ui/src/app/shared/Utils.ts
+++ b/ui/packages/ui/src/app/shared/Utils.ts
@@ -349,7 +349,7 @@ export function minimizePropertyValues (propertyValues: Map<string, string>, pro
       if (propDefn.isMandatory) {
         minimizedValues.set(key, value);
       // Include non-mandatory if has default and value is different
-      }else if (propDefn.defaultValue) {
+      }else if (propDefn.defaultValue != null) {
         if (propDefn.defaultValue !== value) {
           minimizedValues.set(key, value);
         }


ui/packages/ui/src/app/shared/Utils.ts
```diff
@@ -349,7 +349,7 @@ export function minimizePropertyValues (propertyValues: Map<string, string>, pro
       if (propDefn.isMandatory) {
         minimizedValues.set(key, value);
       // Include non-mandatory if has default and value is different
-      }else if (propDefn.defaultValue) {
+      }else if (propDefn.defaultValue != null) {
```
rk3rn3r, line 352: Hey @mdrillin can you maybe have a look into the backend code and fix sending any null to the UI too with that PR?

mdrillin: Sure, but I'd suggest working that as a separate PR.  Is there a JIRA for it?

rk3rn3r: You can just do it here in this PR. The purpose is the same.

mdrillin: The purpose is not the same.  This PR eliminates sending of properties with default values from the UI to the backend.

rk3rn3r: That's a question of the pov. "Fix issue with removal of default properties" can also mean: one side is fixing issues with such data in the frontend, but can also cover eliminating sending the wrong data in the first place from the backend.

mdrillin:
@indraraj @uidoyen - this can be reviewed and merged.  The other issue in question was resolved separately.  DBZ-3967 gives an example of the properties affected by this PR.

indraraj (approved):
Tested locally, looks good.

PR #396, opened 2021-10-01
DBZ-3696 Addition of Topic Groups to UI

mdrillin:
Draft - don't merge.
Addition of topic creation groups to the UI connector wizard


indraraj:
@mdrillin I was thinking if could get some suggestions from UX team for handling this topic creation disabled case  
image

PR #405, opened 2021-11-12
DBZ-4274 Rework usages of replaceAll in frontend

mdrillin:
Encountered issues in firefox browser due to usages of 'replaceAll' javascript function.
Reworked the 'replaceAll' usages to instead use the 'replace' function.


uidoyen (approved):
LGTM!

PR #406, opened 2021-11-16
DBZ-4293 Corrects issue with UI step refresh

mdrillin:
Corrects UI regression in the previous commit - RegEx was being used incorrectly.  After entering topic groups, proceeding to next step, then going back to topic groups step - the connector properties were not being maintained.  This commit removes the RegEx usage in favor of a different approach.


uidoyen (approved):
LGTM!

PR #409, opened 2021-11-19
DBZ-4109 DBZ-4234 Upgrade Debezium and other dependencies

rk3rn3r:
DBZ-4109 Upgrade Debezium and other dependencies (Quarkus, etc) of Debezium UI

closes https://issues.redhat.com/browse/DBZ-4109

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -14,6 +14,7 @@
   <name>Debezium UI Backend</name>
 
   <properties>
+    <quarkus.package.type>uber-jar</quarkus.package.type>
     <logmanager.class>org.jboss.logmanager.LogManager</logmanager.class>
   </properties>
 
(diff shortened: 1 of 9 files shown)


gunnarmorling:
@rk3rn3r, where are we with this one? What's missing in order to wrap it up?

backend/pom.xml
```diff
@@ -14,6 +14,7 @@
   <name>Debezium UI Backend</name>
 
   <properties>
+    <quarkus.package.type>uber-jar</quarkus.package.type>
```
gunnarmorling, line 17: Oh, what's the reason for the uber-JAR? This was consciously using the multi-layer approach before.

gunnarmorling (commented):
The dep updates LGTM, but why are we moving to an uber -JAR? It's generally not a recommended practice for container-based deployments.

backend/pom.xml
rk3rn3r: See https://issues.redhat.com/browse/DBZ-4621

PR #410, opened 2021-11-19
DBZ-4119 Correct enablement of create connector steps

mdrillin:
This addresses the issue raised in DBZ-4119, plus a couple other issues with the create connector wizard:
1) DBZ-4119 - When user validates connection and goes directly to 'Review', now the LH menu 'Properties' step is enabled.
2) The LH menu selection was not always correctly indicated - this PR resolves
3) If user goes back to 'Properties' step and changes a connection property, the previously visited steps are invalidated and their properties removed.  The intent here is to prevent the case where the user can connect to a completely different db, rendering previously entered optional properties invalid (filter, topic group, etc).  They will now be required to revisit and re-enter those optional properties if the connection is changed.


uidoyen (approved):
LGTM!

PR #412, opened 2021-11-29
DBZ-4378 Add required defaults when topic groups are defined

mdrillin:
Ensure that the topic group replication factor and partitions defaults are included whenever topic groups have been defined.  These two defaults are required whenever topic groups are defined, according to table 1 here - https://debezium.io/documentation/reference/configuration/topic-auto-create-config.html#default-topic-creation-group-configuration


gunnarmorling:
@uidoyen, are you looking into this one?

uidoyen (approved):
LGTM!

PR #416, opened 2022-01-06
Setup husky to fix/check code formatting

uidoyen:


ui/package.json
```diff
@@ -11,20 +11,26 @@
...
     "cypress": "cypress run --headless --browser chrome",
+    "format": "lerna run format",
+    "lint-staged": "lerna run lint-staged",
     "unit:test": "lerna run test",
     "i18n": "lerna run i18n --scope=@debezium/ui"
   },
   "husky": {
     "hooks": {
-      "pre-commit": "yarn build"
+      "pre-commit": "yarn run lint-staged"
```
indraraj, line 26: yarn build already have the prebuild script to check for lint. you can instead update the rebuild script with this

uidoyen: I think it's bit confusing. Here https://github.com/debezium/debezium-ui/blob/main/ui/packages/ui/package.json#L12 `prebuild` runs `yarn clean`  and here https://github.com/debezium/debezium-ui/blob/main/ui/package.json#L14  `yarn lint` also It would be enough to just check for formatting and linting issues instead of building the UI. `cos-ui` follows the same https://github.com/bf2fc6cc711aee1a0c2a/cos-ui/blob/main/package.json#L144-L151

indraraj: > I think it's bit confusing. Here https://github.com/debezium/debezium-ui/blob/main/ui/packages/ui/package.json#L12 `prebuild` runs `yarn clean` 

this is for the React UI project that we have inside ui/packages 
> and here https://github.com/debezium/debezium-ui/blob/main/ui/package.json#L14 `yarn lint` 

this is the Lerna cmd which runs `yarn lint` inside all the three sub-projects

uidoyen: Yes, that's the confusion, we should have identical for both level.

indraraj: actually having identical cmd at both levels for this will result in running the same cmd twice in this project. By having different rebuild scripts we are able to run two diff operations sequentially.

uidoyen: I know that but my concern is that why we call `prebuild` at one level to lint the UI and at the second level to clean the UI? We already have https://github.com/debezium/debezium-ui/blob/main/ui/packages/ui/package.json#L20

.github/workflows/frontend-testing.yml
```diff
@@ -40,3 +40,5 @@ jobs:
           working-directory: ./ui
       - name: Test Debezium UI Frontend
         run: yarn run unit:test
+      - name: Format
```
indraraj, line 43: Hey wanted to know why we need this here.

ui/packages/ui/package.json
```diff
@@ -9,17 +9,19 @@
   "license": "Apache-2.0",
   "private": true,
   "scripts": {
-    "prebuild": "yarn clean",
     "build": "webpack --progress --config webpack.prod.js",
     "start": "webpack serve --color --progress --config webpack.dev.js",
     "start:prod": "node scripts/serve",
     "lint": "tslint -c ./tslint.json --project .",
+    "lint:fix": "npm run lint --fix",
```
indraraj, line 16: we should yarn cmd for uniformity

indraraj (approved):
LGTM. Left some comments you can reply to and merge. Also wanted to know if we should set up the 'lint-staged' cmd in the other two packages WDYT?

.github/workflows/frontend-testing.yml
uidoyen: I think it's not required any longer as we are already using with husky.

uidoyen:
> LGTM. Left some comments you can reply to and merge. Also wanted to know if we should set up the 'lint-staged' cmd in the other two packages WDYT?

I think it's a good idea. Let me handle this in another PR.

PR #418, opened 2022-01-12
DBZ-4538 Update connector list immediately upon connector delete

mdrillin:
Refreshes the Debezium UI connector list immediately upon a successful connector deletion.  Previously, the connector list would not update until the next page polling cycle.  Also improved some of the toast messages to include the connector name.


indraraj (approved):
Tested locally, LGTM.

PR #419, opened 2022-01-12
DBZ-4540 Corrects issue with nulls in Mongo connector UI

mdrillin:
Minor change to MongoDbConnectorIntegrator to eliminate null in the namespaces.

Before:
MongoFiltersBefore

After:
MongoFiltersAfter


rk3rn3r (approved):
LGTM @mdrillin! The test error is unrelated, it seems.

PR #437, opened 2022-01-25
DBZ-4460 Extracting version property

gunnarmorling:


ani-sha (approved):

rk3rn3r:
@gunnarmorling @ani-sha That's sad. I handled this already in another PR and I think it's missing that we could better use `${revision]. See other PRs first please!

gunnarmorling:
Well, in that case it would have helped if you had mentioned that fact on the original PR when merging it, or better yet, just push that change to the original PR itself. I don't see how it's sad if I'm fixing left-overs here, seriously.

gunnarmorling:
Hum, I don't see how that is related then. My fix was to that PR which just brought in that duplication; I don't see how this could have been reported or fixed before. But the entire discussion is pretty moot, there was a small issue, I fixed it, let's move on. Why do I need to justify mself for doing that? This doesn't add any value, it only makes me feel miserable.

PR #446, opened 2022-01-25
[CI] Fix CI build issues

rk3rn3r:


pom.xml
```diff
@@ -7,22 +7,26 @@
...
   <groupId>io.debezium</groupId>
   <artifactId>debezium-ui-parent</artifactId>
-  <version>1.0.0-SNAPSHOT</version>
+  <version>${revision}</version>
   <packaging>pom</packaging>
   <name>Debezium UI Build Aggregator</name>
 
   <properties>
+    <revision>1.9.0-SNAPSHOT</revision>
+    <debezium.base.version>${revision}</debezium.base.version>
```
gunnarmorling, line 16: What's the reason for having two separate variables here? I feel "revision" is a bit too generic btw. Can we converge on a single one, named "debezium.version"?

gunnarmorling (approved):
This LGTM overall. One minor question/suggestion inline.

pom.xml
rk3rn3r: @gunnarmorling  The reason is that revision is handled differently than other vars by Maven. It can be used to model dependencies of modules (like can be used in <parent> blocks). See: https://maven.apache.org/maven-ci-friendly.html

rk3rn3r:
@ani-sha @uidoyen @indraraj Could you have a look too?

pom.xml
gunnarmorling: Ok, can't we use `revision` then everywhere? Do you want to leave the door open for having different versions for UI and Debezium itself?

Taking a step back, I'd prefer to keep build set-up as similar as possible between UI and main repo, i.e. so to reduce mental toil (this discussion being the prime example). So instead of using that revision stuff (which is nice in and by itself), I'd just have followed the known path and moved to the revision property in a concerted action for all the repos. No need to change it back here now, though.

gunnarmorling:
LGTM, leaving it to any other review to merge it. Thx, @rk3rn3r!

pom.xml
rk3rn3r: Yes, I thought we might come into some situations around release time where it might be necessary to have temporary different versions.

I would personally suggest that we change the main repo to switch to the `$revision` process too? Should we maybe discuss this in the next team call? I would like to pitch it there. wdyt?

gunnarmorling: Sure; If you wanna go for it, I'd suggest though you send a mail to the (public) list instead.

Personally, I don't exactly see the value. Having that version duplicated in the `parent` blocks of the sub-modules isn't much of a problem, as all this is automatically updated during releases anyways. I think there's many other things on our plates where spending our energy has a much better pay off.

rk3rn3r:
Thx @gunnarmorling! I promise I can't describe how happy I am now that we finally solved the CI errors. :sweat:

gunnarmorling:
Yes, definitely, that's great news!

PR #533, opened 2022-02-16
DBZ-4714: fix: handle the new data_shape key and value

indraraj:


uidoyen (approved):
Tested locally. Looks good!

PR #586, opened 2022-04-29
DBZ-5067: fix: Debezium MCS Error when changing Postgres port

indraraj:
https://issues.redhat.com/browse/DBZ-5067


uidoyen (approved):

indraraj:
> @indraraj I am not sure If we should include MCS in the upstream title. The rest Looks good!

sry @uidoyen didn't get what you meant by "include MCS in the upstream title"

PR #613, opened 2022-05-23
DBZ-5125 Remove spatial_ref_sys table from Postgres filter tests

ani-sha:
https://issues.redhat.com/browse/DBZ-5125

- [x] Fixes failures in `ValidatePostgresFiltersIT` test in reference with https://issues.redhat.com/browse/DBZ-4814
- [x] Fix NPE race condition in several tests.

### DIFF
--- a/backend/src/test/java/io/debezium/configserver/CreateAndDeleteMySqlConnectorIT.java
+++ b/backend/src/test/java/io/debezium/configserver/CreateAndDeleteMySqlConnectorIT.java
@@ -72,8 +72,7 @@ public class CreateAndDeleteMySqlConnectorIT {
                 deleteMySqlConnectorName,
                 Infrastructure.getMySqlConnectorConfiguration(1)
         );
-        Infrastructure.getDebeziumContainer().ensureConnectorTaskState(
-                deleteMySqlConnectorName, 0, Connector.State.RUNNING);
+        Infrastructure.waitForConnectorTaskStatus(deleteMySqlConnectorName, 0, Connector.State.RUNNING);
 
         given()
                 .when().delete(ConnectorURIs.API_PREFIX + ConnectorURIs.MANAGE_CONNECTORS_ENDPOINT, 1, deleteMySqlConnectorName)
--- a/backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
+++ b/backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
@@ -6,9 +6,11 @@
 package io.debezium.configserver.util;
 
 import io.debezium.connector.mongodb.MongoDbConnectorConfig;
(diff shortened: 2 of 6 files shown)


backend/src/test/java/io/debezium/configserver/CreateAndDeleteMySqlConnectorIT.java
```diff
@@ -72,8 +72,7 @@ public void testMySqlDeleteConnectorSuccessful() {
                 deleteMySqlConnectorName,
                 Infrastructure.getMySqlConnectorConfiguration(1)
         );
-        Infrastructure.getDebeziumContainer().ensureConnectorTaskState(
-                deleteMySqlConnectorName, 0, Connector.State.RUNNING);
+        Infrastructure.waitForConnectorTaskStatus(deleteMySqlConnectorName, 0, Connector.State.RUNNING);
```
Naros, line 75: So I elected to add this method to `Infrastructure`. For now this replaces the call to `ensureConnectorTaskState` because the latter introduces potential NPE issues, fixed in DBZ-5159 (as documented in the method comments), but also the ensure-method explicitly uses a timeout that isn't applicable for certain scenarios such as some PostgreSQL failures that are retried with a small delay in-between.  The tests should be resilient to these semantics and that's what this new method is meant to cover.  I adjusted several call locations in the tests to use the logic universally rather than litter the same logic in each test use case.

backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
```diff
@@ -157,7 +160,7 @@ public static ConnectorConfiguration getMySqlConnectorConfiguration(int id, Stri
                 .with("database.server.name", "dbserver" + id)
                 .with("database.history.kafka.bootstrap.servers", KAFKA_HOSTNAME + ":9092")
                 .with("database.history.kafka.topic", "dbhistory.inventory")
-                .with("server.id", "debezium_" + (5555 + id - 1));
+                .with("database.server.id", Long.valueOf(5555 + id - 1));
```
Naros, line 163: There were two problems here.  

* The property should have been `database.server.id`.  This was probably masked previously because the UI and the connector would automatically generate a value on Debezium 1.9.  We now force the UI and Debezium 2.0 to no longer auto-generate a value and so it must be supplied here.
* The value needs to be of type `Long`, not of type `String` as previously provided.  This problem was again masked by the fact we used to auto-generate the value and now we don't via the schema generator changes upstream in the 1.9 branch..

Naros (approved):
Hi @ani-sha, I've left some comments here as an FYI so you understand the reason for these changes.

Naros:
Applied to `main`, thanks @ani-sha.

PR #631, opened 2022-07-22
DBZ-5437 Move to Debezium 2.0.0.Alpha3 as baseline

Naros:
https://issues.redhat.com/browse/DBZ-5437

This addresses two key issues:
* The first is that in Debezium 1.9.3.Final, the signature for the `PostgresConnection` changed and the usage was not updated in the UI repository.  Commit f9722be addresses this by aligning the signature usage.
* The integration tests use `debezium/connect:nightly` which is based on Debezium 2.0 at this point.  As a result the usage of the `database.dbname` has been removed in favor of `database.names`.  I've updated the tests to reflect this change.

This highlights one point of discussion as it relates to the version of Debezium that this project is based upon.  In the POM, it currently references `1.9.0-SNAPSHOT` and yet our tests are explicitly testing against the nightly container build that is based upon 2.0.  

In this PR I have forwarded the Debezium dependency to 2.0.0.Alpha3 explicitly.  I have left the POM version reference still as 1.9.0-SNAPSHOT until we agree on whether we should advance this or not.

### DIFF
--- a/.github/workflows/backend-testing.yml
+++ b/.github/workflows/backend-testing.yml
@@ -22,10 +22,11 @@ jobs:
     steps:
       - name: Checkout Debezium UI
         uses: actions/checkout@v2
-      - name: Set up JDK 11
+      - name: Set up JDK 17
         uses: actions/setup-java@v1
         with:
-          java-version: 11
+          distribution: 'temurin'
+          java-version: 17
       - name: Cache local Maven repository
         uses: actions/cache@v2
         with:
--- a/backend/src/main/java/io/debezium/configserver/service/postgres/PostgresConnectorIntegrator.java
+++ b/backend/src/main/java/io/debezium/configserver/service/postgres/PostgresConnectorIntegrator.java
@@ -113,7 +113,7 @@ public class PostgresConnectorIntegrator extends ConnectorIntegratorBase {
         additionalMetadata.put(PostgresConnectorConfig.PROPAGATE_COLUMN_SOURCE_TYPE.name(), new AdditionalPropertyMetadata(false, ConnectorProperty.Category.CONNECTOR_ADVANCED));
         additionalMetadata.put(PostgresConnectorConfig.MSG_KEY_COLUMNS.name(), new AdditionalPropertyMetadata(false, ConnectorProperty.Category.CONNECTOR_ADVANCED));
         additionalMetadata.put(PostgresConnectorConfig.INCLUDE_UNKNOWN_DATATYPES.name(), new AdditionalPropertyMetadata(false, ConnectorProperty.Category.CONNECTOR_ADVANCED));
-        additionalMetadata.put(PostgresConnectorConfig.TOASTED_VALUE_PLACEHOLDER.name(), new AdditionalPropertyMetadata(false, ConnectorProperty.Category.CONNECTOR_ADVANCED));
+        additionalMetadata.put(PostgresConnectorConfig.UNAVAILABLE_VALUE_PLACEHOLDER.name(), new AdditionalPropertyMetadata(false, ConnectorProperty.Category.CONNECTOR_ADVANCED));
         additionalMetadata.put(PostgresConnectorConfig.PROVIDE_TRANSACTION_METADATA.name(), new AdditionalPropertyMetadata(false, ConnectorProperty.Category.CONNECTOR_ADVANCED));
         additionalMetadata.put(PostgresConnectorConfig.SANITIZE_FIELD_NAMES.name(), new AdditionalPropertyMetadata(false, ConnectorProperty.Category.CONNECTOR_ADVANCED));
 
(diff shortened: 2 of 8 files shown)


pom.xml
```diff
@@ -13,7 +13,11 @@
 
   <properties>
     <revision>1.9.0-SNAPSHOT</revision>
-    <debezium.base.version>${revision}</debezium.base.version>
+    <!--
+    For now lets explicitly lock into a specific Debezium version
+    Do not forget to adjust .github/workflows/end-to-end-testing.yaml to align with this version
+    -->
+    <debezium.base.version>2.0.0.Alpha3</debezium.base.version>
```
rk3rn3r, line 20: This should be done only after @ani-sha's PR to update to 1.9.4. Also I think updating the `$revision` is the right way to do the upgrade or is there a reason to only update the base version?

ani-sha:
@Naros LGTM! One thing to point out is that we have https://issues.redhat.com/browse/DBZ-5317 for upgrading UI to 1.9.5.Final and then set up separate branches it 1.9 and 2.0 which I was working on currently.

backend/src/main/java/io/debezium/configserver/service/postgres/PostgresConnectorIntegrator.java
```diff
@@ -138,7 +137,7 @@ public FilterValidationResult validateFilters(Map<String, String> properties) {
 
         PostgresConnectorConfig config = new PostgresConnectorConfig(Configuration.from(properties));
 
-        try (PostgresConnection connection = new PostgresConnection(config.getJdbcConfig())) {
+        try (PostgresConnection connection = new PostgresConnection(config.getJdbcConfig(), PostgresConnection.CONNECTION_GENERAL)) {
```
rk3rn3r, line 140: This should be handled by @ani-sha for [DBZ-5317](https://issues.redhat.com/browse/DBZ-5317) not here.

pom.xml
ani-sha: @rk3rn3r Should I open my PR first or should we create the branches first? I am confused about that part.

.github/workflows/backend-testing.yml
```diff
@@ -22,10 +22,11 @@ jobs:
     steps:
       - name: Checkout Debezium UI
         uses: actions/checkout@v2
-      - name: Set up JDK 11
+      - name: Set up JDK 17
         uses: actions/setup-java@v1
         with:
-          java-version: 11
+          distribution: 'temurin'
```
rk3rn3r, line 28: why `temurin`? Don't we prefer vanilla `adopt`/`adopt-hotspot`?

backend/src/test/java/io/debezium/configserver/CreateAndDeleteSqlSeverConnectorIT.java
```diff
@@ -74,7 +74,7 @@ public void testSqlServerDeleteConnectorFailed() {
     @Test
     public void testSqlServerDeleteConnectorSuccessful() {
         ConnectorConfiguration config = Infrastructure.getSqlServerConnectorConfiguration(1)
-                .with("database.dbname", "testdb");
+                .with("database.names", "testdb");
```
rk3rn3r, line 77: nightly was only needed in the past to get in a change from 1.9.0 which was not yet released. We can go back with the [DBZ-5317](https://issues.redhat.com/browse/DBZ-5317) work to a stable version instead of nightly. Then switch to `2.0.0.Alpha3` images once we have 1.9 branch decoupled. wdyt?

backend/src/test/java/io/debezium/configserver/CreateAndDeleteSqlSeverConnectorIT.java
```diff
@@ -46,7 +46,7 @@ public void testSqlServerClustersEndpoint() {
     @Test
     public void testSqlServerCreateConnectorEndpoint() {
         ConnectorConfiguration config = Infrastructure.getSqlServerConnectorConfiguration(1)
-                .with("database.dbname", "testdb");
+                .with("database.names", "testdb");
```
rk3rn3r, line 49: see below/above

pom.xml
rk3rn3r: @ani-sha I would personally do all 1.9 related work on the main branch first. Then branch out a 1.9 branch. This way we don't need to merge back 1.9 branch commits to the main branch, resulting in cleaner history imo and it might be easier for other contributors. But if you prefer another way that's also doable.

Naros: As I commented on the 1.9.5.Final PR, we need to expressly use the `debezium/connect:1.9` branch for that work and we need to address any concerns about whether all the bits for the REST extension are there, etc.  It makes no sense to forward the unit tests and the source to 1.9.5.Final but to perform integration tests against 2.0.

backend/src/main/java/io/debezium/configserver/service/postgres/PostgresConnectorIntegrator.java
Naros: That's fine, I can rebase against `main` after her work is addressed.

.github/workflows/backend-testing.yml
Naros: It's the same used in the core repository.

backend/src/test/java/io/debezium/configserver/CreateAndDeleteSqlSeverConnectorIT.java
Naros: I think for `main`, we should explicitly test against `nightly`.  The whole point of those integration tests being executed in the UI action within the core repository in the first place is to try and identify PR failures in the core repository that have an impact on the UI.  If we explicitly test against say 2.0.0.Alpha3, then any future changes to core will not reflect any impacts and we'll be doing work after-the-fact rather than catching the problems when the code changes.

Naros:
Closed in favor of #635.

PR #635, opened 2022-08-01
DBZ-5317 Upgrade UI to 2.0

ani-sha:
https://issues.redhat.com/browse/DBZ-5317

### DIFF
--- a/.github/workflows/backend-testing.yml
+++ b/.github/workflows/backend-testing.yml
@@ -22,10 +22,11 @@ jobs:
     steps:
       - name: Checkout Debezium UI
         uses: actions/checkout@v2
-      - name: Set up JDK 11
-        uses: actions/setup-java@v1
+      - name: Set up JDK 17
+        uses: actions/setup-java@v2
         with:
-          java-version: 11
+          distribution: 'temurin'
+          java-version: 17
       - name: Cache local Maven repository
         uses: actions/cache@v2
         with:
--- a/.github/workflows/end-to-end-testing.yml
+++ b/.github/workflows/end-to-end-testing.yml
@@ -26,17 +26,17 @@ jobs:
 
     services:
       dbzui-zookeeper:
-        image: debezium/zookeeper:1.6
+        image: debezium/zookeeper:2.0
 
       dbzui-kafka:
-        image: debezium/kafka:1.6
+        image: debezium/kafka:2.0
         ports:
           - 9092:9092
         env: 
           ZOOKEEPER_CONNECT: dbzui-zookeeper:2181
 
       dbzui-postgres:
-        image: debezium/example-postgres:1.6
+        image: debezium/example-postgres:2.0
         ports:
           - 65432:5432
         env: 
--- a/.github/workflows/end-to-end-testing.yml
+++ b/.github/workflows/end-to-end-testing.yml
@@ -26,17 +26,17 @@ jobs:
 
     services:
       dbzui-zookeeper:
-        image: debezium/zookeeper:2.0
+        image: debezium/zookeeper:latest
 
       dbzui-kafka:
-        image: debezium/kafka:2.0
+        image: debezium/kafka:latest
         ports:
           - 9092:9092
         env: 
           ZOOKEEPER_CONNECT: dbzui-zookeeper:2181
 
       dbzui-postgres:
-        image: debezium/example-postgres:2.0
+        image: debezium/example-postgres:latest
         ports:
           - 65432:5432
         env: 
--- a/backend/src/test/java/io/debezium/configserver/ValidateSqlSeverDatabasesIT.java
+++ b/backend/src/test/java/io/debezium/configserver/ValidateSqlSeverDatabasesIT.java
@@ -14,7 +14,6 @@ import io.quarkus.test.junit.QuarkusTest;
 import io.quarkus.test.junit.TestProfile;
 import io.restassured.http.ContentType;
 
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
 import java.util.Map;
@@ -29,7 +28,6 @@ import static org.hamcrest.CoreMatchers.hasItems;
 public class ValidateSqlSeverDatabasesIT {
 
     @Test
-    @Disabled("Requires upgrading to latest Debezium 1.9 to fix bugs with newly added database.names property")
     public void testMultipleDatabases() {
         ObjectNode config = ConnectorConfigurationTestingHelper.getConfig(
                 Infrastructure.getSqlServerConnectorConfiguration(1)
--- a/backend/src/test/java/io/debezium/configserver/ValidateSqlSeverDatabasesIT.java
+++ b/backend/src/test/java/io/debezium/configserver/ValidateSqlSeverDatabasesIT.java
@@ -5,23 +5,19 @@
  */
 package io.debezium.configserver;
 
-import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.debezium.configserver.rest.ConnectorURIs;
 import io.debezium.configserver.util.Infrastructure;
 import io.debezium.configserver.util.SqlServerInfrastructureTestProfile;
-import io.debezium.testing.testcontainers.ConnectorConfigurationTestingHelper;
+import io.debezium.testing.testcontainers.Connector;
+import io.debezium.testing.testcontainers.ConnectorConfiguration;
 import io.quarkus.test.junit.QuarkusTest;
 import io.quarkus.test.junit.TestProfile;
 import io.restassured.http.ContentType;
 
 import org.junit.jupiter.api.Test;
 
-import java.util.Map;
-
 import static io.restassured.RestAssured.given;
 import static org.hamcrest.CoreMatchers.equalTo;
-import static org.hamcrest.CoreMatchers.is;
-import static org.hamcrest.CoreMatchers.hasItems;
 
 @QuarkusTest
 @TestProfile(SqlServerInfrastructureTestProfile.class)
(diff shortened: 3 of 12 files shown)


Naros:
Hi @ani-sha so I looked at the failures, and yes you will need to include a dependency on:
```
<dependency>
  <groupId>io.debezium</groupId>
  <artifactId>debezium-storage-kafka</artifactId>
</dependency>
````

File: backend/src/test/java/io/debezium/configserver/ValidateSqlSeverDatabasesIT.java [hunk 1]
```diff
@@ -5,48 +5,39 @@
...
     @Test
     public void testMultipleDatabases() {
-        ObjectNode config = ConnectorConfigurationTestingHelper.getConfig(
-                Infrastructure.getSqlServerConnectorConfiguration(1)
-                        .with("database.hostname", "localhost")
-                        .with("database.port", Infrastructure.getSqlServerContainer().getMappedPort(1433))
-                        .with("database.names", "testdb,testdb2")
-        );
+        ConnectorConfiguration config = Infrastructure.getSqlServerConnectorConfiguration(1)
+                .with("database.names", "master,model");
```
ani-sha, line 29: @Naros @rk3rn3r I was enabling this test for 2.0 to support multiple databases and I am getting this error. 
```
{
    "error_code": 400,
    "message": "Connector configuration is invalid and contains the following 1 error(s):\nOnly a single database name is currently supported\nYou can also find the above list of errors at the endpoint `/connector-plugins/{connectorType}/config/validate`"
}
```

Naros: That's definitely a strange error and not what I see.  When I run the test suite I actually get
```
{
    "error_code": 400,
    "message": "Connector configuration is invalid and contains the following 1 error(s):\nUnable to connect. Check this and other connection properties. Error: The database 'master' is not enabled for Change Data Capture. Ensure that the correct database context is set and retry the operation. To report on the databases enabled for Change Data Capture, query the is_cdc_enabled column in the sys.databases catalog view.\nYou can also find the above list of errors at the endpoint `/connector-plugins/{connectorType}/config/validate`"
}
```
This makes sense because you cannot enabled CDC on `master` and this check was in fact added as a part of DBZ-4346.  I believe you need to readjust the databases to be `testdb,testdb2` since that is what we initialize the database with.

Naros: That's definitely a strange error and not what I see.  When I run the test suite I actually get
```
{
    "error_code": 400,
    "message": "Connector configuration is invalid and contains the following 1 error(s):\nUnable to connect. Check this and other connection properties. Error: The database 'master' is not enabled for Change Data Capture. Ensure that the correct database context is set and retry the operation. To report on the databases enabled for Change Data Capture, query the is_cdc_enabled column in the sys.databases catalog view.\nYou can also find the above list of errors at the endpoint `/connector-plugins/{connectorType}/config/validate`"
}
```
This makes sense because you cannot enabled CDC on `master` and this check was in fact added as a part of DBZ-4346.  I believe you need to readjust the databases to be `testdb,testdb2` since that is what we initialize the database with.

ani-sha: @Naros Did the tests pass locally for you? This time the error 
```
{
    "error_code": 400,
    "message": "Connector configuration is invalid and contains the following 1 error(s):\nUnable to connect. Check this and other connection properties. Error: The database 'testDB2' is not enabled for Change Data Capture. Ensure that the correct database context is set and retry the operation. To report on the databases enabled for Change Data Capture, query the is_cdc_enabled column in the sys.databases catalog view.\nYou can also find the above list of errors at the endpoint `/connector-plugins/{connectorType}/config/validate`"
}
```

.github/workflows/end-to-end-testing.yml
```diff
@@ -26,17 +26,17 @@ jobs:
 
     services:
       dbzui-zookeeper:
-        image: debezium/zookeeper:1.6
+        image: debezium/zookeeper:2.0
```
Naros, line 29: Should this target `latest` rather than `2.0` specifically?

.github/workflows/end-to-end-testing.yml
```diff
@@ -26,17 +26,17 @@ jobs:
...
 
       dbzui-kafka:
-        image: debezium/kafka:1.6
+        image: debezium/kafka:2.0
```
Naros, line 32: Should this target `latest` rather than `2.0` specifically?

.github/workflows/end-to-end-testing.yml
```diff
@@ -26,17 +26,17 @@ jobs:
...
         ports:
           - 9092:9092
         env: 
           ZOOKEEPER_CONNECT: dbzui-zookeeper:2181
 
       dbzui-postgres:
-        image: debezium/example-postgres:1.6
+        image: debezium/example-postgres:2.0
```
Naros, line 39: Should this target `latest` rather than `2.0` specifically?

.github/workflows/backend-testing.yml
```diff
@@ -22,10 +22,11 @@ jobs:
...
       - name: Checkout Debezium UI
         uses: actions/checkout@v2
-      - name: Set up JDK 11
-        uses: actions/setup-java@v1
+      - name: Set up JDK 17
+        uses: actions/setup-java@v2
         with:
-          java-version: 11
+          distribution: 'temurin'
+          java-version: 17
```
Naros, line 29: Don't we also want to check out the Debezium core repository like we do in the other connector repositories and make sure we compile the latest source from upstream and store those artifacts before testing Debezium UI?

Naros (approved):
Hi @ani-sha this LGTM overall, just a few comments inline and I think we're good.

.github/workflows/backend-testing.yml
rk3rn3r: Didn't we agree to explicitly depend on 2.0 instead of latest until we manually update? ` latest` might block too many PRs until breaking changes from the main repo are resolved?

rk3rn3r:
Didn't we agree to explicitly depend on 2.0 instead of latest until we manually update? ` latest` might block too many PRs until breaking changes from the main repo are resolved?

.github/workflows/backend-testing.yml
ani-sha: @Naros I intend to do this on a separate PR.

Naros:
But explicitly depending on `2.0` would also mean that running the upstream UI check that uses the `nightly` image might also cause the same problem because there may be changes in the nightly build that aren't yet in the published `2.0` images.

ani-sha:
> The `2.0` to `latest` were targetting images that rarely change, zookeeper, kafka, and the example databases. In addition, the `latest` and `2.0` are synonymous tags anyway, but it just helps minimize the upgrade tasks when we move from 2.0 to 2.1, etc. I'd prefer if we can minimize all the effort involved when new releases go out, that just helps all involved.

I think we can give this a try. If things don't work that well we can switch back to `2.0` tag anytime. Wdyt @rk3rn3r?

rk3rn3r:
@ani-sha I don't know if I am wrong here. Let' me try to explain why I feel unsure about using `latest`: When I look at it from the UI side of things depending on 1.9/2.0/2.1 explicitly (even better more explicitly on `1.9.4`, `1.9.5`, `2.0.0.Beta1`) feels more stable and reliable and even reproducible. I can always be sure that a newly released `latest` version won't break tests / PRs immediately.
Instead I need to create a manual PR to update to the next latest version and I can fix issues with the PR, I can even ignore broken releases.
For example this is the reason I think is what's causing troubles with the core repo tests: core changes are breaking UI which is breaking core because those 2 repos are not well integrated (because they aren't mono repo) and I don't see that we are able to keep up with the core change rate to keep everything stable. Even more: in the future we need to be able to handle multiple Debezium versions with the UI.

Please correct me when I am wrong, or if you find this is an acceptable situation, then I agree using `latest`. But if my assumptions are correct I personally prefer the safer and explicit dependency, but I am okay with using `latest` when you want to keep it now and see how it goes.

ani-sha:
@rk3rn3r I understand your concern. Then I guess there aren't any changes to make in this PR. Could you go ahead and merge?

rk3rn3r:
It depends. I would personally make the dependent versions more explicit (`2.0.0.Beta1`) for example. Chris is more on the `latest` side. It's up to you how you want to go. Let's wait for @Naros to allay my concerns maybe later.

Naros:
So to be clear there are two aspects as it pertains to versions.

**Image versions in the end-to-end workflow**
This is specifically what I think we're referring to, the tag for `debezium/kafka`, `debezium/zookeeper`, and the databases.  It does not matter whether we use `2.0` or `latest` in this context, they're the same SHA.  The only benefit to using `latest` is that when we move from one maintenance release to another, there isn't a need to make any changes in the UI repository.

Furthermore, looking to the future with multiple Debezium versions, we would replace `latest` with a matrix of versions anyway, and in this matrix `latest` would be one of those.  Technically this job could be adjusted to using a matrix build pattern that only uses `latest` and we can just add more to it in the future when we get to that point.

**Explicit micro-releases vs Snapshot**
So we have 3 workflows, two in the UI repository, and one in the main repository.  The one in the main repository based the core repository based on the current state of `main` as a snapshot and builds the UI and then runs its tests.  Ideally, this job should run the UI against the current Debezium version, which is `2.0.0-SNAPSHOT` to guarantee that we identify if there are any compatibility problems between the core and the UI or vice versa.

If we pin the Debezium UI's `pom.xml` to a specific version, such as `2.0.0.Beta1`, then that workflow in the core repository is a waste because it's not verifying anything in the context of the core repository at all.  We'll never see if a PR to the core repo causes any type of incompatibility because it'll always pass.  That's why I mentioned that if we went this route, we should just remove the UI step in the core repository entirely because it serves no value.

This also means that in terms of keeping the UI up-to-date, that will fall out of the purview of the contributors to the core or even those of us who merge changes into the core as we'll have no indicator to tell us that such incompatibility exists.

I think if there is a reason to be sticky to a given version, you could leverage a `2.0` branch where that branch is confirmed to be aligned with `2.0.0.Beta1` and you could use this branch as a means of tracking the sticky versions but allowing the `main` branch to be fluid so we gain the benefit of keeping it up-to-date as changes in core occur in parallel.

rk3rn3r:
Applied. Thx a lot @ani-sha.

PR #640, opened 2022-08-26
DBZ-5229 Restore Kafka-based database history options

Naros:
https://issues.redhat.com/browse/DBZ-5229

This follows a similar principle as done in the schema generator in the core repository in PR https://github.com/debezium/debezium/pull/3829 by which we explicitly inject the Kafka-based connector fields into the configuration if the connector contains a `database.history` property as a part of its base configuration definition.

This approach doesn't support other storage modules, but it at least restores the broken functionality until a larger discussion can be had on how best to integrate these additional storage modules with the UI project and downstream components.


rk3rn3r (approved):
LGTM. Thx a lot @Naros 

Is this SQL Server error in the CI test related?

Naros:
It is, I'll go ahead and merge this @rk3rn3r.  Thanks for the review.

PR #653, opened 2022-09-15
DBZ-5613 Wait before SQL server container is considered as started

vjuranek:
We execute init script once the container is started, but started container may not be initialized yet and e.g. setting CDC for tables may fail with:

    Cannot perform this operation while SQLServerAgent is starting. Try again later.

This happens especially in CI where the machines are slower. Add startup strategy to wait for 5 seconds until container is considered as started.

https://issues.redhat.com/browse/DBZ-5613


jpechane:
@vjuranek Applied, thanks

PR #656, opened 2022-09-22
DBZ-5585 DBZ-5496 Classloader no longer needed; Azure SQL compatibility

jpechane:
- [ ] Changes related https://github.com/debezium/debezium/pull/3858 not yet present


Naros:
Hi @jpechane and @vjuranek, it would seem that the internal CI job isn't publishing docker images due to this compilation problem.  How do we want to handle this so we can release 1.9.6.Final container images?  Presently we haven't published any images it seems for the past week.

vjuranek (approved):
LGTM, also tests are passing locally for me.

vjuranek:
> How do we want to handle this so we can release 1.9.6.Final container images?

remove UI image from the 1.9.6 release and release 1.9.7 ASAP?

jpechane:
@vjuranek Thanks for verification!

jpechane:
The image is available now - there is `SKIP_UI` env var for scripts to ignore UI image. So hotfix is done. Just for this cases we should extend the job to have this env var available as parameter as I had to add it directly to the pipeline

PR #657, opened 2022-09-29
DBZ-5664 Fix ValidatePostgresConnectionIT.testInvalidPostgresConnecti…

vjuranek:
…on test

After removing default value for `topic.prefix` there is one more error message in the response.

https://issues.redhat.com/browse/DBZ-5664


jpechane:
@vjuranek Applied, thanks for the fast turnaround.

PR #676, opened 2022-11-22
DBZ-5865 Add connector.displayName and connector.id to Config endpoint

Naros:
https://issues.redhat.com/browse/DBZ-5865

This adds the `connector.displayName` and `connector.id` to the config endpoint payload.  
@uidoyen let me know if this needs to be adjusted.


uidoyen (approved):
@Naros thanks for including this :+1:

PR #693, opened 2023-03-03
DBZ-6173 Upgrade to Debezium 2.2 & align configuration properties

ani-sha:
https://issues.redhat.com/browse/DBZ-6173


backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
```diff
@@ -40,8 +40,7 @@ public class Infrastructure {
     public enum DATABASE {
         POSTGRES, MYSQL, SQLSERVER, MONGODB, ORACLE, NONE
     }
-
-    private static final String DEBEZIUM_CONTAINER_VERSION = "2.0";
+    private static final String DEBEZIUM_CONTAINER_VERSION = "2.1";
```
rk3rn3r, line 43: This should be 2.2 as well.

backend/src/main/java/io/debezium/configserver/service/mongodb/MongoDbConnectorIntegrator.java
```diff
@@ -99,9 +96,9 @@ public class MongoDbConnectorIntegrator extends ConnectorIntegratorBase {
         MONGODB_PROPERTIES =  Collections.unmodifiableMap(additionalMetadata);
     }
 
-    protected ConnectionContext.MongoPrimary primary(MongoDbTaskContext context) throws Throwable {
-        ReplicaSet replicaSet = ReplicaSet.parse(context.getConnectionContext().hosts());
-        return context.getConnectionContext().primaryFor(replicaSet, context.filters(), (s, throwable) -> {
+    protected MongoDbConnection primary(MongoDbTaskContext context) throws Throwable {
```
rk3rn3r, line 99: HOSTS and AUTO_DISCOVER_MEMBERS above are deprecated options. Should we remove them?  Also below in tests, should we better provide the `CONNECTION_STRING = Field.create("mongodb.connection.string")` instead?

backend/src/main/java/io/debezium/configserver/service/mongodb/MongoDbConnectorIntegrator.java
```diff
@@ -99,9 +96,9 @@ public class MongoDbConnectorIntegrator extends ConnectorIntegratorBase {
...
+        ReplicaSet replicaSet = new ReplicaSet(context.getConnectionContext().connectionString());
```
rk3rn3r, line 100: I think the test is failing as the new Mongo DB instance is using a mapped port which needs to be fetched, could that be? Not entirely sure.

backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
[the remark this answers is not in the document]
ani-sha: @rk3rn3r What mongodb version should we test in the UI?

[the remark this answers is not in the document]
rk3rn3r: @ani-sha wdym? Debezium Example images are aligned with the major Debezium version, e.g. `quay.io/debezium/example-mongodb:2.2`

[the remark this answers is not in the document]
Naros: I think MongoDB 5 is a good middle-ground.  We certainly do not want to go any lower than MongoDB 4.2 IIRC.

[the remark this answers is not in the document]
jcechace: 4.2 is going to be deprecated soon as well. So 5.0 is indeed a great middle ground between 4.4 and 6.0

Naros:
Hi @ani-sha I looked at the MongoDB container start-up error and I have made two observations.

First, MongoDB changed how they log the "start-up" log message with MongoDB 4.4.  In older versions the string they were looking for was `"waiting for connections"`; however since MongoDB 4.4 this was changed to a log entry of `"Waiting for connections"`, notice the capital `W`.

At first, I thought this could be resolved by updating test containers, but then I realized in the project's root pom.xml, we are specifying Testcontainers with a version of 1.17.6 and importing the bom; however, this has no impact on the version being used in the 'debezium-ui-backend` project, it continues to use 1.15.2.  Even when explicitly setting all the test containers dependencies with the version of `${testcontainers.version}` and verifying that the project was now using the latest 1.17.6 version; this didn't affect the outcome of the failure.

So first I would say we need to address this version issue with the POM, as I would have expected setting the test containers version in the root pom of the project would have translated to the testcontainers dependency used in the child modules, but that did not.

Now, to address the start-up problem with the MongoDB container, I then modified the `Infrastructure` class wit this small code for the definition of the MongoDB container:

```
    private static final MongoDBContainer MONGODB_CONTAINER =
            new MongoDbContainer(DockerImageName.parse("mongo:5.0"))
                    .withNetwork(NETWORK)
                    .withNetworkAliases("mongodb")
                    .waitingFor(Wait.forLogMessage(".*Waiting for connections*.*", 1));
```

This successfully starts the MongoDB 5.0 container, but unfortunately this leads to another problem with authentication which I then solved by amending the above as follows:

```
    private static final MongoDBContainer MONGODB_CONTAINER =
            new MongoDbContainer(DockerImageName.parse("mongo:5.0"))
                    .withNetwork(NETWORK)
                    .withNetworkAliases("mongodb")
                    .waitingFor(Wait.forLogMessage(".*Waiting for connections*.*", 1))
                    .withCommand("--replSet", "rs0");
```

Both of the above problems stem from our `MongoDbContainer` implementation in the upstream repository, which we need to discuss with @jcechace because the wait message only works for older variants of MongoDB and the `--auth` command used by the wrapper doesn't seem to work with MongoDB 5 unless there is some configuration or set up we aren't doing in the UI project.

rk3rn3r:
Like I have mentioned in a previous reply, we should switch to using the upstream `MongoDbContainer` from the `debezium-testing-testcontainers` package. But it needs some enhancements before we can use it. It would be great if we could support:
a) using an image with pre-populated data/collections and users, like `quay.io/debezium/example-mongodb:2.2` (done by @Naros recently)
b) enabling startup init script usage to populate data on container creation, like we do at the moment: https://github.com/debezium/debezium-ui/blob/8e96f28f81498d0747fa792b9515dc0f4c758768/backend/src/test/java/io/debezium/configserver/util/MongoDbContainer.java#L41 , which copies the init file from the resources directory.

backend/pom.xml
```diff
@@ -142,6 +142,7 @@
     <dependency>
       <groupId>org.testcontainers</groupId>
       <artifactId>testcontainers</artifactId>
+      <version>${testcontainers.version}</version>
```
rk3rn3r, line 145: This is not required. Testcontainers dependency is linked from the parent pom's `<dependencyManagement>` section using the testcontainers BOM.

Naros: So the culprit here is the inclusion of the `quarkus-bom`, which also imports the `testcontainers-bom`.  The version of Quarkus being used in the backend imports 1.15.2, which is why that version is used there.  

I tried testing exclusion of the `testcontainers-bom` when we import the `quarkus-bom`, but this does not seem to work and we still get the 1.15.2 dependencies.  What does work however is if we import the `testcontainers-bom` inside the backend pom.xml but prior to the `quarkus-bom` and all testcontainer dependencies then follow our specified testcontainers.version.  

Steps:
1. Remove `testcontainers-bom` from the room pom.xml
2. Adjust the backend pom.xml:
```
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>org.testcontainers</groupId>
        <artifactId>testcontainers-bom</artifactId>
        <version>${testcontainers.version}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
      <dependency>
        <groupId>${quarkus.platform.group-id}</groupId>
        <artifactId>${quarkus.platform.artifact-id}</artifactId>
        <version>${quarkus.platform.version}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
  ```
  3. The dependency tree is now:
  ```
  └─$ ./mvnw dependency:tree | grep testcontainers
[INFO] +- org.testcontainers:testcontainers:jar:1.17.6:test
[INFO] +- org.testcontainers:junit-jupiter:jar:1.17.6:test
[INFO] +- io.debezium:debezium-testing-testcontainers:jar:2.2.0-SNAPSHOT:compile
[INFO] +- org.testcontainers:kafka:jar:1.17.6:compile
[INFO] +- org.testcontainers:jdbc:jar:1.17.6:compile
[INFO] |  \- org.testcontainers:database-commons:jar:1.17.6:compile
[INFO] +- org.testcontainers:postgresql:jar:1.17.6:test
[INFO] +- org.testcontainers:mysql:jar:1.17.6:test
[INFO] +- org.testcontainers:mongodb:jar:1.17.6:test
[INFO] \- org.testcontainers:mssqlserver:jar:1.17.6:test
```

Naros: Can we all agree on this, because without using 1.17.6, we have container issues and so we must explicitly upgrade to this version somehow, either by explicitly overriding the version in all the artifacts listed in the backend pom or by importing the `testcontainers-bom` prior to the `quarkus-bom` within the same maven pom as shown above.

rk3rn3r: Yes, that makes sense. Maybe we should put a detailed comment in the backend pom there to explain why it must be there, if order is important, etc.

ani-sha: @Naros @rk3rn3r Added a brief comment and updated the pom files accordingly.

backend/pom.xml
```diff
@@ -182,6 +190,12 @@
       <artifactId>mssqlserver</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>net.java.dev.jna</groupId>
+      <artifactId>jna</artifactId>
```
rk3rn3r, line 195: Out of curiosity, what is this used for?

backend/pom.xml
```diff
@@ -182,6 +190,12 @@
...
+      <version>5.12.1</version>
```
rk3rn3r, line 196: Can we define the version in the parent pom's `DependencyManagement` section like we do with the other dependencies?

backend/pom.xml
ani-sha: The JNA dependency was needed for my local setup, isn't actually needed in the UI code. So I will remove it.

rk3rn3r:
Could you please update all docker compose files too?

backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
```diff
@@ -65,10 +65,19 @@ public enum DATABASE {
...
+        MONGODB_CONTAINER = MongoDbContainer.node()
+                .name("localhost")
+                .replicaSet("rs0")
+                .imageName(DockerImageName.parse("mongo:5.0"))
+                .network(NETWORK)
+                .build();
+        MONGODB_CONTAINER
+                .withNetworkAliases("mongodb")
+                .withCopyFileToContainer(MountableFile.forClasspathResource("initialize-mongo-single.js"), "/docker-entrypoint-initdb.d/")
+                .initReplicaSet(false, MONGODB_CONTAINER.getNamedAddress());
```
rk3rn3r, line 79:
```suggestion
                .build()
                .withNetworkAliases("mongodb")
                .withCopyFileToContainer(MountableFile.forClasspathResource("initialize-mongo-single.js"), "/docker-entrypoint-initdb.d/");
                MONGODB_CONTAINER.initReplicaSet(false, MONGODB_CONTAINER.getNamedAddress());
```

backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
```diff
@@ -40,8 +41,7 @@ public class Infrastructure {
     public enum DATABASE {
         POSTGRES, MYSQL, SQLSERVER, MONGODB, ORACLE, NONE
     }
-
```
rk3rn3r, line 43: Shouldn't we keep that empty line?

backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
rk3rn3r: And shouldn't the `MONGODB_CONTAINER.initReplicaSet(false, MONGODB_CONTAINER.getNamedAddress());` command go to the `io.debezium.configserver.util.MongoDbInfrastructureTestResourceLifecycleManager#start` method? A replica set can only be initialized for a started container, right?

Naros: This won't work because these methods require a started container, I've delayed this until after the container has started.

backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
Naros: Added back.

Naros:
Per discussions with Jiri, I've temporarily disabled all MongoDB tests, allowing us to proceed with the upgrade until we can find a working solution.  The issue with MongoDB is that the current API doesn't work well for allowing connections from both the inside and outside wrt how we're using the container implementation, therefore we agreed to disable the tests for the time being.  If CI passes, I'll merge and I'll add a follow-up task.

Naros (approved):
The test passes locally, so in the interest of the release, I will merge the PR.

rk3rn3r:
Just for the record: That was not my preferred solution.
Do you think we should re-evaluate the tight coupling of the core and UI release and dependency?

PR #697, opened 2023-03-30
DBZ-5313:Edit a connector in Debezium UI

uidoyen:


indraraj:
While randomly clicking on the side nav I got this error, screenshot below 
image
maybe this will help

indraraj (changes requested):
added two comments

indraraj:
also WDYT about demoing it in our UI call?

uidoyen:
> While randomly clicking on the side nav I got this error, screenshot below image maybe this will help

It should be fine now.

indraraj:
@uidoyen  If you want you can get this in since its an extensive PR, and for any further improvement we can follow up in subsequent PR's

PR #702, opened 2023-04-20
DBZ-5359 Connector metrics API endpoint implementation

ani-sha:
https://issues.redhat.com/browse/DBZ-5359

Reference output : https://gist.github.com/ani-sha/e41b7132f6a82c2a43d1264a3fae4490 (individual responses)


Naros:
Hi @ani-sha, it would appear we may need to define something similar to `KafkaConnectClientFactory` as we cannot rely on the `localhost:8778` setting in the `application.properties` when working with test containers.

backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
```diff
@@ -96,7 +96,10 @@ public enum DATABASE {
                 .withNetwork(NETWORK)
                 .withKafka(KAFKA_CONTAINER.getNetwork(), KAFKA_HOSTNAME + ":9092")
                 .withLogConsumer(new Slf4jLogConsumer(LOGGER))
-                .dependsOn(KAFKA_CONTAINER);
+                .dependsOn(KAFKA_CONTAINER)
+                .withStartupTimeout(Duration.ofSeconds(120))
+                .withEnv("ENABLE_JOLOKIA", "true")
+                .withExposedPorts(8083, 8778);
```
rk3rn3r, line 102: @Naros  Shouldn't this be added to a DebeziumContainer builder in the core repo? something like `#enableJolokia()` there?

backend/src/main/java/io/debezium/configserver/rest/client/JolokiaClient.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Inject
+    JolokiaAttributes jolokiaAttributes;
+
+    public List<String> getAttributeNames() {
+        return jolokiaAttributes.getAttributeNames();
+    }
+
+    public List<J4pReadResponse> getMetrics(String connectorType, String serverName, List<String> attributeNames) {
+        final J4pClient client = getJolokiaClient();
```
rk3rn3r, line 38: Instead of having a proxy for Jolokia, why don't we access and expose the metrics directly? If Jolokia is enabled for the Debezium Connect container then the UI could use Jolokia directly, why using this proxy?

ani-sha:
> it would appear we may need to define something similar to KafkaConnectClientFactory as we cannot rely on the localhost:8778 setting in the application.properties when working with test containers.

Hi @Naros, But why can't we rely on `application.properties`? It is fetching the `jolokia.client.uri` but it is being treated as unknown host by jolokia.

backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
Naros: That seems reasonable since Jolokia is baked in.

backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
```diff
@@ -96,7 +96,10 @@ public enum DATABASE {
                 .withNetwork(NETWORK)
                 .withKafka(KAFKA_CONTAINER.getNetwork(), KAFKA_HOSTNAME + ":9092")
                 .withLogConsumer(new Slf4jLogConsumer(LOGGER))
-                .dependsOn(KAFKA_CONTAINER);
+                .dependsOn(KAFKA_CONTAINER)
+                .withStartupTimeout(Duration.ofSeconds(120))
```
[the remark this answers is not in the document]
Naros: We do, but that code block in `Infrastructure` you're referencing only applies if the `CI` environment variable is set.  Can you confirm whether we need this timeout when running locally or in a non-CI environment?

backend/src/main/java/io/debezium/configserver/rest/client/JolokiaClient.java
Naros: > Instead of building this proxy for Jolokia, why don't we access and expose the metrics directly via Jolokia without an endpoint?

Are you referring to the `ConnectorURIs.LIST_CONNECTOR_METRICS_ENDPOINT` endpoint or something else?  Perhaps we can sync up on this when you're back from PTO?

backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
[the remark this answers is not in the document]
rk3rn3r: Previously we didn't have timeout related issues on local machines, but on CI.

backend/src/main/java/io/debezium/configserver/rest/client/JolokiaClient.java
rk3rn3r: Yes, let's sync up tomorrow when I'm back.
What I meant was that we don't build that endpoint inside the backend but instead we only enable Jolokia and let the UI Frontend access it directly...? But we should think about security constraints, sure.

indraraj:
@ani-sha @Naros we should be able to merge this after we finalized to use custom-backed API for exposing Jolokia metrics in today's team call. WDYT?

ani-sha:
@indraraj Yes, I would leave it to @Naros for a final review and merge.

backend/src/main/java/io/debezium/configserver/rest/client/JolokiaClient.java
rk3rn3r: I suggest we discuss this in another call and we need to reconsider how the UI frontend accesses the metrics. I think there's currently no good way than to access Jolokia directly when taking UI backend removal into account. Similarly to how we agreed on in the milestone planning to reduce effort.

ani-sha:
> I don't think we should have merged that PR.

@rk3rn3r which PR are you talking about?

backend/src/main/java/io/debezium/configserver/rest/client/JolokiaClient.java
Naros: We already discussed this with the broader team on June 8th's team call and we all concluded that the current design made sense.  We can certainly rediscuss this again, but I think we need to also consider this PR is nearly 2 months old and we need to move on to other tasks too.

rk3rn3r:
A big sorry. It looks like I mixed up the bunch of emails when I was catching up on all the things after my PTO and sick leave. 
Please forget all that I was saying about that topic. :facepalm:

ani-sha:
@Naros @rk3rn3r Hi, can we get this merged? So that it would unblock @indraraj in the UI work! Thanks.

backend/src/main/resources/application.properties
```diff
@@ -10,6 +10,9 @@ quarkus.http.test-ssl-port=0
 # only enable the next line in order to allow CORS request from any domain for all modes (dev and prod, etc), e.g. for UI frontend development (see details in README.md), when you can't use the dev mode
 # quarkus.http.cors.origins=*
 
+# Jolokia endpoint inside Kafka Connect images
+jolokia.client.uri=http://dbzui-connect:8778/jolokia
```
rk3rn3r, line 14: I  don't think this is correct. Shouldn't the hostname be derived from the connect cluster (based on the `{cluster}` URL parameter)? The only option should be the port name. Maybe we should extend the Kafka Connect cluster config for this instead?

backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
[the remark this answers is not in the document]
rk3rn3r: Can you remove this and see if it works? There are extended timeouts set for CI.

rk3rn3r:
@ani-sha Can you have a look at my previous and some new comments that I have just added?

backend/src/main/resources/application.properties
ani-sha: @rk3rn3r Applied the changes, but the testcontainers test seems to fail. It works fine on the UI.

rk3rn3r: Hey @ani-sha! Great to see progress here! Let me have a look

backend/src/main/java/io/debezium/configserver/rest/client/JolokiaClient.java
```diff
@@ -0,0 +1,62 @@
...
+import org.jolokia.client.J4pClient;
+import org.jolokia.client.exception.J4pException;
+import org.jolokia.client.request.J4pReadRequest;
+import org.jolokia.client.request.J4pReadResponse;
+
+@ApplicationScoped
+public class JolokiaClient {
+    public static final int DEFAULT_JOLOKIA_PORT = 8778;
+
+    public static final String PROPERTY_JOLOKIA_CLIENT_URI = "jolokia.client.uri";
```
rk3rn3r, line 26:
```
Caused by: jakarta.enterprise.inject.spi.DeploymentException: io.quarkus.runtime.configuration.ConfigurationException: Failed to load config value of type class java.lang.String for: jolokia.client.uri
```
Is causing the test failure. I think you need to remove this one and all usages.

backend/src/main/java/io/debezium/configserver/rest/client/JolokiaClient.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @ConfigProperty(name = PROPERTY_JOLOKIA_CLIENT_URI)
```
rk3rn3r, line 28: And this one.

backend/src/main/java/io/debezium/configserver/rest/client/JolokiaClient.java
```diff
@@ -0,0 +1,62 @@
...
+    private String configJolokiaUrl;
+    @Inject
+    JolokiaAttributes jolokiaAttributes;
```
rk3rn3r, line 31: This is also related to my previous comment about injection. Constructor or setter injection should make it easier to mock and/or test this code. @Naros @jpechane wdyt?

backend/src/main/resources/application.properties
rk3rn3r: @ani-sha I have looked at the CI tests and commented below what might be causing the issue.

backend/src/test/java/io/debezium/configserver/util/AbstractInfrastructureTestResourceLifecycleManager.java
```diff
@@ -20,6 +21,10 @@ public Map<String, String> start() {
                 KafkaConnectClientFactory.PROPERTY_KAFKA_CONNECT_URIS,
                 "http://" + Infrastructure.getDebeziumContainer().getHost() + ":" + Infrastructure.getDebeziumContainer().getMappedPort(8083)
         );
+        config.put(
+                JolokiaClient.PROPERTY_JOLOKIA_CLIENT_URI,
```
ani-sha, line 25: @rk3rn3r If we remove the config property how can we pass the URL into our connector config?

rk3rn3r: I think you switched the main code to use the URI of the selected Kafka Connect cluster (based on the `{{cluster}}` URL parameter? This shouldn't be necessary now, is it? I feel confused now, am I wrong?

ani-sha: Since we are using the testcontainers, we need to use the DebeziumContainer's host and mapped port like we do in line 20 here for `kafka.connect.uri`. Otherwise it doesn't work and throws this error :
```
HTTP Request to /api/connectors/1/my-postgres-connector/metrics failed, error id: 5ab3b6c8-5940-4aa2-b9e2-53630fb30504-1: org.jboss.resteasy.spi.UnhandledException: java.lang.RuntimeException: org.jolokia.client.exception.J4pConnectException: Cannot connect to http://localhost:8778/jolokia: Connect to localhost:8778 [localhost/127.0.0.1, localhost/0:0:0:0:0:0:0:1] failed: Connection refused (Connection refused)
	at org.jboss.resteasy.core.ExceptionHandler.handleApplicationException(ExceptionHandler.java:110)
	at org.jboss.resteasy.core.ExceptionHandler.handleException(ExceptionHandler.java:370)
	at org.jboss.resteasy.core.SynchronousDispatcher.writeException(SynchronousDispatcher.java:218)
	at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:519)
	at org.jboss.resteasy.core.SynchronousDispatcher.lambda$invoke$4(SynchronousDispatcher.java:261)
	at org.jboss.resteasy.core.SynchronousDispatcher.lambda$preprocess$0(SynchronousDispatcher.java:161)
	at org.jboss.resteasy.core.interception.jaxrs.PreMatchContainerRequestContext.filter(PreMatchContainerRequestContext.java:364)
	at org.jboss.resteasy.core.SynchronousDispatcher.preprocess(SynchronousDispatcher.java:164)
	at org.jboss.resteasy.core.SynchronousDispatcher.invoke(SynchronousDispatcher.java:247)
	at io.quarkus.resteasy.runtime.standalone.RequestDispatcher.service(RequestDispatcher.java:82)
	at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler.dispatch(VertxRequestHandler.java:147)
	at io.quarkus.resteasy.runtime.standalone.VertxRequestHandler$1.run(VertxRequestHandler.java:93)
	at io.quarkus.vertx.core.runtime.VertxCoreRecorder$14.runWith(VertxCoreRecorder.java:576)
	at org.jboss.threads.EnhancedQueueExecutor$Task.run(EnhancedQueueExecutor.java:2513)
	at org.jboss.threads.EnhancedQueueExecutor$ThreadBody.run(EnhancedQueueExecutor.java:1538)
	at org.jboss.threads.DelegatingRunnable.run(DelegatingRunnable.java:29)
	at org.jboss.threads.ThreadLocalResettingRunnable.run(ThreadLocalResettingRunnable.java:29)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: java.lang.RuntimeException: org.jolokia.client.exception.J4pConnectException: Cannot connect to http://localhost:8778/jolokia: Connect to localhost:8778 [localhost/127.0.0.1, localhost/0:0:0:0:0:0:0:1] failed: Connection refused (Connection refused)
```

rk3rn3r: @ani-sha I was suggesting on Google Chat that we do a quick call to settle on this.
The only thing that we should be using is https://github.com/debezium/debezium-ui/blob/main/backend/src/test/java/io/debezium/configserver/util/AbstractInfrastructureTestResourceLifecycleManager.java#L19-L22 and we are now using the rewritten URI from the related cluster from the `{{cluster}}` parameter, as we do now here:
https://github.com/debezium/debezium-ui/pull/702/files#diff-b0c80e985e9cdbb002cbbb501120be53968a781acacce6e9bbe55476a333838cR548-R549

backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
```diff
@@ -56,6 +61,9 @@ public class ConnectorResource {
 
     private static final Logger LOGGER = Logger.getLogger(ConnectorResource.class);
 
+    @Inject
+    JolokiaClient jolokiaClient;
```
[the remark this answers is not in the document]
Naros: If we were talking about some other dependency injection frameworks, perhaps, but Quarkus simplifies this to the point that the choice you make here is irrelevant as you can use the same test injection style regardless if the bean uses `@Inject`, constructor, or setter -based injection:

```
@BeforeAll
public static void applyJolokiaClientMockForTests() {
    JolokiaClient client = Mockito.mock(JolokiaClient.class);
    // Mockito lines to build the client stub we want for tests
    QuarkusMock.installMockForType(client, JolokiaClient.class)
}
```

I think since we've mostly (if not completely used) the style @ani-sha used here, I'd favor consistency since it really doesn't impact the testing approach we'd need to use for Quarkus.

backend/src/main/java/io/debezium/configserver/rest/client/JolokiaClient.java
Naros: See my earlier comment about `QuarkusMock` and how this is really irrelevant as their style works the same regardless of the method used, making building tests a breeze for this kind of thing.

backend/src/main/java/io/debezium/configserver/rest/client/JolokiaClient.java
Naros: So I suggest we try and be flexible in this regard.

I see in `ConnectoResource` line 549, we have this logic:
```
String.format("http://%s:%d/jolokia", kafkaConnectURI.getHost(), JolokiaClient.DEFAULT_JOLOKIA_PORT);
```

I think this is a great way to handle the default situation where the user may be using Jolokia bundled with our container images; however, this may prove problematic for other installations where the agent may have been installed specifying an `agentContext` other than the default of `jolokia` or using a non-default port.

So I think there is merit in allowing users to override this URI in the application properties if they have added the Jolokia client in a way that deviates from using the default agent configuration values.  In order to support this, the change is quite simple to make the code that Anisha has already in place work for both.

First, you can make a `@ConfigProperty` binding not be required by wrapping the variable as an `Optional`, and this will allow Quarkus to safely start without failure:
```
    @ConfigProperty(name = PROPERTY_JOLOKIA_CLIENT_URI)
    private Optional<String> configJolokiaUrl;
```
Then I'd suggest modifying line 39 of the `JolokiaClient` to:
```
        final J4pClient client = getJolokiaClient(configJolokiaUrl.orElse(jolokiaUrl));
```
Running the test suite with this modification works as expected regardless of whether the Jolokia endpoint is constructed from the hardcoded template in the resource or by being specified in application properties.

Naros:
I've left some inline comments @ani-sha and @rk3rn3r and I think that covers all the issues in this PR.  I've manually ran the tests locally with my suggested fixes and the tests all pass:
```
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Debezium UI Build Aggregator 2.3.0-SNAPSHOT:
[INFO] 
[INFO] Debezium UI Build Aggregator ....................... SUCCESS [  0.091 s]
[INFO] Debezium UI Frontend ............................... SUCCESS [ 47.742 s]
[INFO] Debezium UI Backend ................................ SUCCESS [02:41 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  03:30 min
[INFO] Finished at: 2023-06-28T11:48:44-04:00
[INFO] ------------------------------------------------------------------------
```
Let me know if you have any other questions.

backend/src/main/java/io/debezium/configserver/rest/client/JolokiaClient.java
rk3rn3r: @Naros  How is this going to work with users operating multiple Connect clusters from a single UI?

rk3rn3r: > I think this is a great way to handle the default situation where the user may be using Jolokia bundled with our container images;

I think in this phase it is okay to say this is the only supported way to run UI, using our images or configure Jolokia in a compatible way. When we move to the final solution  based on the Debezium Connect REST Extension this all will go away anyway, right?

Naros: Would it be reasonable @rk3rn3r and @ani-sha to treat the `jolokia.client.uri` property in the same way we do the Kafka Connect URI property, that if multiple values are provided for Kafka Connect that we expect a similar comma-separated list of URIs for Jolokia?

rk3rn3r: It would, but I think the effort is to high considering that we are going to move to another solution when migrating to the Debezium Connect REST Extension. Let's just rewrite the Kafka Connect URIs that come from the `{{cluster}}` URL parameter. Imo it is a good enough assumption to enable the feature and finally have it out of the door (with those standard config limitations). We can do things properly when we move it to the REST extension.

backend/src/main/java/io/debezium/configserver/rest/client/jolokia/JolokiaClient.java
```diff
@@ -0,0 +1,92 @@
...
+import org.jolokia.client.J4pClient;
+import org.jolokia.client.exception.J4pException;
+import org.jolokia.client.request.J4pReadRequest;
+import org.jolokia.client.request.J4pReadResponse;
+
+@ApplicationScoped
+public class JolokiaClient {
+    public static final int DEFAULT_JOLOKIA_PORT = 8778;
+    public static final String PROPERTY_JOLOKIA_CLIENT_URI = "jolokia.client.uri";
+    private static JolokiaURIsList jolokiaURIsList;
```
rk3rn3r, line 34: Why are we not using List<URI> here instead of wrapping ArrayList with a custom data type?

backend/src/main/java/io/debezium/configserver/rest/client/jolokia/JolokiaClient.java
```diff
@@ -0,0 +1,92 @@
...
+            try {
+                return getMetrics(client, connectorType, serverName, attributeName);
+            }
+            catch (MalformedObjectNameException | J4pException e) {
+                throw new RuntimeException(e);
+            }
+        }).collect(Collectors.toList());
+    }
+
+    public J4pReadResponse getMetrics(J4pClient client, String connectorType, String serverName, String attributeName)
```
rk3rn3r, line 79: This one could be private? The client is instantiated only internally, or?

backend/src/main/java/io/debezium/configserver/rest/client/jolokia/JolokiaClient.java
```diff
@@ -0,0 +1,92 @@
...
+import org.jolokia.client.exception.J4pException;
+import org.jolokia.client.request.J4pReadRequest;
+import org.jolokia.client.request.J4pReadResponse;
+
+@ApplicationScoped
+public class JolokiaClient {
+    public static final int DEFAULT_JOLOKIA_PORT = 8778;
+    public static final String PROPERTY_JOLOKIA_CLIENT_URI = "jolokia.client.uri";
+    private static JolokiaURIsList jolokiaURIsList;
+    @ConfigProperty(name = PROPERTY_JOLOKIA_CLIENT_URI)
```
rk3rn3r, line 35: This should be removed, no?

backend/src/main/java/io/debezium/configserver/model/JolokiaURIsList.java
```diff
@@ -0,0 +1,29 @@
...
+
+import org.eclipse.microprofile.openapi.annotations.enums.SchemaType;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.Collection;
+
+@Schema(implementation = String.class, type = SchemaType.ARRAY, example = "[\"http://localhost:1234\"]")
+public class JolokiaURIsList extends ArrayList<URI> {
```
rk3rn3r, line 16: Are you sure this is needed? A similar construct in another part of the code was only required for the `@Schema` annotation. This should not be needed in this case. A simple `List<URI>` should do well.

backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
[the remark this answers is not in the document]
rk3rn3r: Hey, I thought about this over the weekend. While this is easy to test with Quarkus, I think in general such kind of injection is often seen as a code smell and maybe we can simply move it to the constructor. This won't hurt, does it?

backend/src/main/java/io/debezium/configserver/rest/client/jolokia/JolokiaClient.java
```diff
@@ -0,0 +1,92 @@
...
+    private Optional<String> configJolokiaUrl;
+    @Inject
+    JolokiaAttributes jolokiaAttributes;
+    public static JolokiaURIsList getJolokiaBaseUris() {
+        if (jolokiaURIsList == null) {
+            List<String> jolokiaBaseUris = ((SmallRyeConfig) ConfigProvider.getConfig())
+                    .getValues(PROPERTY_JOLOKIA_CLIENT_URI, String.class, ArrayList::new);
```
rk3rn3r, line 42: This way it is hard to relate a Kafka Connect cluster to a Jolokia URI. The Kafka Connect cluster config should accept two URIs: the KC URI and a optional second Jolokia URI. When this one is not set, default/fallback to the Kafka Connect URI and rewrite the HTTP port to the default Jolokia port and the path to the default path.

This is another reason why I think doing it like this (and doing it now) is over-engineered and takes more and more time to finish this task. And we already know the code is about to be removed again soon and replaced with something new.

backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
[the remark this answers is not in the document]
rk3rn3r: @jcechace What do people learn at university nowadays? Is this good or bad?

backend/src/main/java/io/debezium/configserver/rest/client/jolokia/JolokiaAttributes.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.apache.commons.io.IOUtils;
+
+/**
+ * A simple cache configurable Jolokia attributes that are to be fetched for metrics requests.
+ *
+ * @author Chris Cranford
+ */
+@ApplicationScoped
+public class JolokiaAttributes {
```
rk3rn3r, line 24: `IoUtil.readLines(IoUtil.getResourceAsStream("attributes.txt", null, getClass(), null, null), StandardCharsets.UTF_8);`

backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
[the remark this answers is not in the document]
mfvitale: I agree with @jcechace more or less. 

But let me add more on this sentence `Additionally with higher amount of dependencies the constructor parameters can often become become chaotic.`

This is exactly why it's better to use constructor injection (a part from framework agnostic), it show you , with the increase of parameters, that probably you class is doing to much stuff. Using attribute injection this usually easy less evident. For example some static code analysis has a built-in rule for checking number of parameters on constructors (https://rules.sonarsource.com/java/RSPEC-107/).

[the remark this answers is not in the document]
rk3rn3r: Yes. That's exactly what I had in mind.
While practically high-level languages with frameworks like Quarkus compile and rewrite your code to proxie classes and you can access and replace basically everything... or alternatively simpler by cracking up hidden members with reflection, it's not only not elegant, it is considered bad for all those reasons.
Thanks for the confirmation all. I didn't meant to be too picky on it and wanted your opinion and experiences. 

@ani-sha in that light, could you move this to the constructor? I think you can still let Quarkus inject the instance for you like in section [5.4. Simplified Constructor Injection](https://quarkus.io/guides/cdi-reference#simplified-constructor-injection).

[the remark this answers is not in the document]
jcechace: > I agree with @jcechace more or less.
> 
> But let me add more on this sentence `Additionally with higher amount of dependencies the constructor parameters can often become become chaotic.`
> 
> This is exactly why it's better to use constructor injection (a part from framework agnostic), it show you , with the increase of parameters, that probably you class is doing to much stuff. Using attribute injection this usually easy less evident. For example some static code analysis has a built-in rule for checking number of parameters on constructors (https://rules.sonarsource.com/java/RSPEC-107/).

While true we have to account for java's verbose naming customs. In practice sometimes you just have the 3-5 parameters which do make a complete sense yet the method signature is already bloated -- and we have that in debezium code base as well.

rk3rn3r:
Just fixed the code-style issue. Seems to work now @ani-sha. Can you confirm it's working locally too?

backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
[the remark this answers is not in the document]
rk3rn3r: @ani-sha can you fix this, please? Move the injected dependency into the constructor, please.

PR #706, opened 2023-05-01
DBZ-6406 Update to Debezium 2.3.x

Naros:
https://issues.redhat.com/browse/DBZ-6406


Naros:
The two failures are related to the fact we have some `debezium/example-xxxx:2.3` artifacts that are not published, such as the PostgreSQL and MySQL example databases.  These are not currently published due to a script failure.

PR #710, opened 2023-05-09
DBZ-6426 Correctly ignore unsupported connectors, including unsupported Debezium connectors

rk3rn3r:
closes https://issues.redhat.com/browse/DBZ-6426

### DIFF
--- a/backend/pom.xml
+++ b/backend/pom.xml
@@ -132,7 +132,7 @@
     <dependency>
       <groupId>org.apache.kafka</groupId>
       <artifactId>connect-api</artifactId>
-      <version>3.2.3</version>
+      <version>3.4.0</version>
     </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
--- a/backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
+++ b/backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
@@ -8,11 +8,9 @@ package io.debezium.configserver.rest;
 import java.io.IOException;
 import java.net.URI;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.ServiceLoader;
 
 import io.debezium.configserver.model.ConnectConnectorConfigResponse;
 import io.debezium.configserver.model.ConnectorStatus;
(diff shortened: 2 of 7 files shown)


backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
```diff
@@ -50,23 +48,14 @@
 import io.debezium.configserver.rest.model.BadRequestResponse;
 import io.debezium.configserver.rest.model.ServerError;
 import io.debezium.configserver.service.ConnectorIntegrator;
+import static io.debezium.configserver.service.ConnectorIntegratorBase.integrators;
+import static io.debezium.configserver.service.ConnectorIntegratorBase.supportedConnectorClassnames;
```
Naros, line 52: Hi @rk3rn3r, one thing I noticed is we are inconsistent with our placement of static imports in the UI code base, some are at the end of the import block, others at the start.  I noticed the majority are toward the end, and that is inconsistent with the code formatting used by the upstream core repository.

We should probably make sure we're following the same code format rules as the upstream repository to make it easier for IDE integrations to use the same formatting rules IMO, would you not agree?

backend/pom.xml
```diff
@@ -132,7 +132,7 @@
     <dependency>
       <groupId>org.apache.kafka</groupId>
       <artifactId>connect-api</artifactId>
-      <version>3.2.3</version>
+      <version>3.4.0</version>
     </dependency>
```
Naros, line 136: I'm curious if we shouldn't base this on the upstream debezium bom?

rk3rn3r: Yes, but before we were diverging as we were behind the latest version or because of RHOC specifics.
Now we could make use of the debezium-bom. Could you file a Jira task for that?

backend/src/main/java/io/debezium/configserver/rest/ConnectorResource.java
rk3rn3r: We are using the core repo checkstyle.xml imo. Does this not cover this? I think aligning it makes sense, although I find there are more important things than sorting imports. ;)

Naros: > We are using the core repo checkstyle.xml imo. Does this not cover this?

That's why I asked the question because in the upstream `debezium-parent.xml` we define this:
```
                <plugin>
                    <groupId>net.revelc.code</groupId>
                    <artifactId>impsort-maven-plugin</artifactId>
                    <version>${version.impsort}</version>
                    <configuration>
                        <groups>java.,jakarta.,javax.,org.,com.,io.</groups>
                        <staticGroups>*</staticGroups>
                        <staticAfter>false</staticAfter>
                        <skip>${format.skip}</skip>
                        <removeUnused>true</removeUnused>
                    </configuration>
                </plugin>
```
So if the UI allows it to build without enforcing static imports at the top, then no the configuration isn't being applied for some reason.  If not, could you log a Jira issue so we don't loose sight of it?

rk3rn3r: Yes, I think this is "missing".

PR #741, opened 2023-06-09
DBZ-6514: build fix

indraraj:


ui/package.json
```diff
@@ -54,7 +54,7 @@
     "copy-webpack-plugin": "11.0.0",
     "css-loader": "6.8.1",
     "css-minimizer-webpack-plugin": "5.0.0",
-    "cypress": "12.14.0",
+    "cypress": "12.13.0",
```
rk3rn3r, line 57: Why do we need this downgrade?

indraraj: since this was a recent update before the build issue appeared I thought of giving it a try.

PR #759, opened 2023-07-06
DBZ-6639 Use Debezium version from pom.xml properties for Testcontainers CI tests

rk3rn3r:
closes https://issues.redhat.com/browse/DBZ-6639

### DIFF
--- a/backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
+++ b/backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
@@ -35,19 +35,34 @@ import java.util.stream.Stream;
 
 public class Infrastructure {
 
+    // Those versions are managed in the debezium-ui-parent pom.xml!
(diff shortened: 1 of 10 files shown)


backend/src/test/java/io/debezium/configserver/util/Infrastructure.java
```diff
@@ -89,7 +104,7 @@ public enum DATABASE {
 
     private static final DebeziumContainer DEBEZIUM_CONTAINER;
     static {
-        DEBEZIUM_CONTAINER = new DebeziumContainer("quay.io/debezium/connect:nightly")
+        DEBEZIUM_CONTAINER = new DebeziumContainer("quay.io/debezium/connect:" + DEBEZIUM_CONTAINER_VERSION)
```
rk3rn3r, line 107: It's really hard to say if we should do this...?!

PR #857, opened 2024-03-04
Fix the Nginx issue when refresh the page

yahyaAlsaidi:
Issue:
When users refresh the page in Nginx, they encounter a "route not found" error due to the default file being served. This occurs because Nginx is configured to serve the default file instead of routing requests to the correct endpoints.

Solution:
To address this issue, I removed the default file from the Nginx configuration and added a custom file to handle route navigation. This ensures that requests are properly routed to the correct endpoints, preventing the "route not found" error on page refresh.


indraraj:
@rk3rn3r can you have a look

nginx.config/default.conf
```diff
@@ -0,0 +1,10 @@
+server {
+    listen 80;
+
+    location / {
+        proxy_buffering off;
+        root /usr/share/nginx/html;
+        try_files $uri $uri/ $uri/index.html /index.html =404;
```
rk3rn3r, line 7: @indraraj how should 404s be handled?

Dockerfile
```diff
@@ -20,6 +20,9 @@ FROM nginx:1.25.3-alpine
 
 ENV KAFKA_CONNECT_CLUSTERS=http://localhost:8083/
 
+RUN rm -rf /etc/nginx/conf.d/*.conf
+COPY ./nginx.config/default.conf /etc/nginx/conf.d/defualt.conf
```
rk3rn3r, line 24: @yahyaAlsaidi could you rename the `nginx.config/` dir to `etc/nginx/` or at least to `etc/`?

nginx.config/default.conf
```diff
@@ -0,0 +1,10 @@
...
+    }
+}
+
```
rk3rn3r, line 10: Default nginx `default.conf` file looks like that:
```
server {
    listen       80;
    server_name  localhost;

    #access_log  /var/log/nginx/host.access.log  main;

    location / {
        root   /usr/share/nginx/html;
        index  index.html index.htm;
    }
    error_page   500 502 503 504  /50x.html;
    location = /50x.html {
        root   /usr/share/nginx/html;
    }

}
```
What is different and really required in your version?

I suppose when there is no `proxy_pass` or similar config proxy is not active? Does this option affect non-proxy requests?

`root` is the same config.
`try_files` is the only difference. I think the use-case is when you are in a sub-page (e.g. `/edit/connector` or something) and the URI in the browser isn't `/` anymore and you then reload the page?
@indraraj is it okay to have a full URI in the browser and should traffic be re-directed to `index.html` then or would we move the URI behind an anchor (e.g. `#/edit/connector`)? Sorry, currently I am not aware of the URIs that we use in the UI frontend.
5xx error config is missing in this version.

The only difference I could spot is try_files and I see a potential use-case for it.
Instead of changing the original `default.conf` maybe we can just add another `conf.d/debezium-ui.conf` to add missing options?

Could you also file a DBZ task under https://issues.redhat.com/ describing the issue and prefix the commit with this?
@ani-sha @indraraj we need to recover the previous checks on PRs. I think it got lost with the removal of the UI backend.

rk3rn3r:
First, @yahyaAlsaidi thx a lot for raising this issue. I left one bigger comment though and I am looking forward to your feedback. Also, like mentioned at the end of my bigger comment, please file a Jira for this change in the Debezium issue tracker / Jira and use this to prefix your commit. And please squash that merge into your actual change-commit. Thx!

Dockerfile
yahyaAlsaidi: Thanks will do, but what the  reason behind make the `nginx.config/` dir renamed to `/etc/nginx/`?

rk3rn3r: @yahyaAlsaidi it's only a convention that more and more projects use to store config files, similar to how Linux stores config. That way it's also easy to understand where those files belong.

nginx.config/default.conf
yahyaAlsaidi: When a user navigates to a specific route in your React application, the browser sends a request to the server for that route. Since Nginx doesn't have a corresponding file or directory for that route (because it's a single-page application), it would normally return a 404 error.

However, by using the `try_files` directive, you can instruct Nginx to always serve the main HTML file (which typically contains the JavaScript bundle for your React application) for any requested URL. This allows your React application to handle client-side routing internally, based on the URL, without needing to make additional requests to the server for each route. For importing the old config and add them for the new one there is no reason behind it but both options we can consider it as okay.

Dockerfile
yahyaAlsaidi: Ah, got it! Now it makes sense. I knew it was a convention, but I was waiting for your reply to get a clearer understanding of it. Thanks for clarifying!

nginx.config/default.conf
yahyaAlsaidi: Will remove it since you don't have 404 html file in the `dist` dir.

indraraj: its a  single page application we don't need a separate 404 html, 404 scenarios are already handled in UI via routes

Dockerfile
rk3rn3r: You are welcome and thx a lot for your effort! :pray:

nginx.config/default.conf
rk3rn3r: Yep, gotcha! That's how I understood your change. So `try_files` is the only addition, right? I wonder what is the correct way to customize (the default) config of nginx for this. I think the best way is adding a new file like I suggested above `conf.d/debezium-ui.conf`. But I need to figure out what needs to be in there. Would be great if you could figure this out @yahyaAlsaidi? Or what do you think?

indraraj: @rk3rn3r from server side we just load the single js bundle on the main domain and the subsequent route are handled on the client side in the SPA react application.

we have to perform some work in modifying the remaining previous checks(like E2E) that we had earlier and for them to work after update. I will log a jira issue(if its not already there) then we can check in the team call if or when we need to pick them.

yahyaAlsaidi: Creating new files is beneficial when managing complex setups with multiple server blocks and numerous location directives. In such cases, organizing configurations into separate files and then including them in the default file can enhance clarity and ease of management. However, in scenarios where there is only one server block and a limited number of location directives to handle, adding a new file might unnecessarily complicate the configuration process. It could lead to confusion and make it more challenging to understand and follow the configuration. In simpler setups like this, directly modifying the existing configuration file is often more straightforward and easier to comprehend. @rk3rn3r If you wish me to do it for you I can do it with only adding the include command in the nginx file here is way of doing it https://serverfault.com/questions/707955/nginx-split-large-configuration-file you can check this one if you want me to do it ?

rk3rn3r: Thx @yahyaAlsaidi for the exploration. I understand and agree to that statement when looking at a specific server's config. My question was a bit more towards dependency management from the UI Nginx container to the upstream Nginx base container. I don't want to be forced to update the default.conf when it was updated upstream. Instead I thought we just keep the original file inherited from the base repo and we only add our additional bits in a new file.
That way the base config is well separated from the UI specific one.
I hope you understand my concern?

nginx.config/default.conf
rk3rn3r: @yahyaAlsaidi I know there is the original `50x.html` file. But I am not sure about 404. I need to check.

nginx.config/default.conf
rk3rn3r: @indraraj but what happens on reload after you navigated to a sub-page/URL? The frontend app that can only be loaded via `index.html` in the current nginx configuration will not be loaded from a sub-URL, right? Then we need this change that @yahyaAlsaidi suggest with this PR in order to correctly load the UI.
BTW, @yahyaAlsaidi @indraraj, the updated nginx config will load the UI and render the correct sub-page that is referenced by the URL?

rk3rn3r:
Sorry @yahyaAlsaidi  for the intensive discussion, but you raised a very good issue and that lead to some things that we should clarify when moving forward.

nginx.config/default.conf
yahyaAlsaidi: Yes, this fix we let you reload the page with the sub-page that is referenced by the URL.

nginx.config/default.conf
yahyaAlsaidi: No, I delete it this part when merging the two commits together, you don't have to check it there is not 404 file there. Sometimes some bundler include that and some not so you don't have it, so there is no reason to add it.

rk3rn3r: Ah yes, I double-checked. There's indeed no 404.html file and the UI frontend is supposed to handle 404s.

One thing that was coming to my mind and I am interested to hear your opinion!
What do you think about using a different directory inside the container to host the UI?
Like `/srv/` or a new directory like `dbz-ui` instead of  `/usr/share/nginx/html`? And then it makes really sense to create a new config file in `conf.d/` dir for that endpoint with your config? That way we done interfere with the default config but we need to try if this will be working. wdyt?

yahyaAlsaidi: Well, simply including your custom configuration in the default file might not resolve the issue due to potential conflicts. Some opt for a different approach by organizing configurations into sites-available and sites-enabled directories. Configurations are placed in sites-available and then symlinked in sites-enabled, which can be included in nginx.conf. However, this method can add complexity, especially if accidental symlink or backup files are created. You can find more discussion on this topic [here](https://serverfault.com/questions/527630/difference-in-sites-available-vs-sites-enabled-vs-conf-d-directories-nginx). While separation of concerns is important, sometimes simplicity is preferable. Also [here](https://antonlytvynov.medium.com/nginx-sites-available-sites-enabled-3bd025bc4d25) kind of example to do something similar to that.

nginx.config/default.conf
indraraj: yes, from the server we need to load the same JS build for all the paths, it's in UI that then handles the exact URL present in the browser to load the necessary UI page for it. I have tested it locally this is fixing the reload issue on suburls.

nginx.config/default.conf
rk3rn3r: You should read the second answer in the first link (Stackoverflow). I think that one is very accurate `Best Practice Is conf.d.` and `You should be using /etc/nginx/conf.d, as that's a standard convention, and should work anywhere.` It think this reply is really on point. In a dockerized environment it does not make sense to use `sites-available`/`sites-enable`. I think this pattern only makes sense when you have one big server hosting multiple different websites/customers etc.

So that leaves the question if we should overwrite the upstream `conf.d/default.conf` or add an additional `conf.d/debezium-ui.conf`? I lean towards the second options if `default.conf` and `debezium-ui.conf` dont't interfere and can work together without harm. Maybe we only need to add the `try_files` line inside the correct structure in `debezium-ui.conf`?

yahyaAlsaidi: You could give it a try.

config.json
```diff
@@ -1,3 +1,3 @@
 {
-    "KAFKA_CONNECT_CLUSTERS": "http://localhost:8083"
+  "KAFKA_CONNECT_CLUSTERS": "http://localhost:8083/"
```
rk3rn3r, line 2: @indraraj  Is this config change needed?

